PR #93, opened 2026-05-05
fix: drop tap-to-reveal overlay from recovery-phrase screen (#82)

### LINKED ISSUE #82: Overlay text floats above visible 12-word backup phrase in settings
## Summary

In the active profile settings under **Backup → Show 12-word recovery phrase → Continue with biometrics**, the recovery phrase screen displays a mysterious overlay text floating directly above the two-column list of 12 words. The overlay text is partially unreadable (contains phrases like "...нажмите чтобы показать...") and appears to serve no clear purpose. The 12 words are already fully visible by default, and tapping inside the word block makes the overlay disappear. This creates confusion for the user: the overlay implies that words are hidden or that an action is required, but neither is true.

---

## Steps to Reproduce

2. Navigate to **Settings** → **Active profile**.
3. Tap on **Backup** section.
4. Tap **Show 12-word recovery phrase**.
5. Authenticate using **biometrics** (fingerprint / face) when prompted.
6. Observe the **"Recovery Phrase"** screen.
7. Notice a ghost / overlay text floating **on top** of the two-column word grid.
8. Try to read the overlay — it is garbled, partially unreadable (e.g., "...нажмите чтобы показать...").
9. Tap anywhere inside the block containing the 12 words.
10. Observe that the overlay disappears.
11. The 12 words remain fully visible both before and after the tap — they were never hidden.

---

## Expected Behavior

- The **12-word recovery phrase** should be displayed clearly, without any confusing overlay text.
- If words are meant to be hidden by default, they should be masked (e.g., dots or asterisks) with a clear **"Tap to reveal"** button or eye icon.
- If words are meant to be visible immediately, there should be **no overlay** suggesting otherwise.
- Any instructional text (e.g., "Write down these words") should be placed **above or below** the word grid, not floating on top of it.

---

## Actual Behavior

- A mysterious, semi-readable overlay text floats **directly on top** of the 12-word grid.
- The overlay implies that the user needs to tap to reveal something — but the words are already visible.
- Tapping inside the word area removes the overlay, but the user is left confused about what just happened.
- No clear user benefit or purpose is evident.

---

dev-onym:
Closes #82.

## Summary

- Recovery-phrase reveal screen no longer draws a "Tap to reveal" overlay over the 12-word grid.
- Words are visible immediately after the biometric authentication; Copy + Continue are enabled from the start.
- `Step.Reveal` no longer carries a `revealed` flag, and `tappedReveal()` is gone.

## Why

Per #82, the previous overlay floated over an apparently-visible 12-word grid, confusing users.

Root cause: the screen used `Modifier.blur(...)` to hide the words pre-reveal, but `Modifier.blur` is a no-op below API 31 (Android 12). On Android 8–11, the words rendered fully readable behind a half-translated overlay reading "...нажмите чтобы показать..." — exactly the screenshot.

The previous attempt on this branch (`65cec7d`) swapped `blur` for `alpha = 0f` so the API < 31 path actually hid the words. That fixed the leak, but kept the tap-to-reveal gate — a second hurdle right after a biometric auth that already passed. The issue body's "Expected Behavior" explicitly calls out that with biometric + FLAG_SECURE already in place, the overlay adds no value.

This PR completes the fix by removing the gate entirely:
- `BiometricAuthenticator` (gates `authenticate()`) + `FLAG_SECURE` on the window (set in `MainActivity`) remain the access controls.
- The phrase appears on entering the Reveal step — same UX as iOS post-auth biometrics flows.
- Drops the `tap_to_reveal` string in en + ru.

## Test plan

- [x] `./gradlew compileDebugKotlin compileDebugAndroidTestKotlin` (clean)
- [x] `./gradlew lintDebug` (clean)
- [x] `./gradlew testDebugUnitTest` — 363/363 of the affected suites pass; 16 pre-existing `Room*StoreTest` failures (SQLite JNI in JVM) are unrelated and reproduce on `main`.
- [ ] Instrumented `RecoveryPhraseBackupScreenTest` + `RecoveryPhraseBackupViewModelTest` updated for the new flow — needs an emulator run in CI.
- [ ] Manual: Settings → Active profile → Backup → Show 12-word recovery phrase → biometrics → words visible immediately, no overlay; Copy + Continue enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,8 +528,7 @@ in shape, ported to Compose + StateFlow + AndroidX BiometricPrompt.
 .intro
   │ tappedContinueFromIntro → authenticate()
   ├─► .authFailed(reason) ── dismissedAuthError → .intro
-  └─► .reveal(phrase, revealed: false)
-        │ tappedReveal     → revealed: true
+  └─► .reveal(phrase)
         │ tappedCopyPhrase → clipboard write + 60s auto-clear
         │ tappedContinueFromReveal
         ▼
--- a/app/src/androidTest/kotlin/chat/onym/android/recovery/RecoveryPhraseBackupScreenTest.kt
+++ b/app/src/androidTest/kotlin/chat/onym/android/recovery/RecoveryPhraseBackupScreenTest.kt
@@ -5,8 +5,6 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertIsEnabled
 import androidx.compose.ui.test.assertIsNotDisplayed
-import androidx.compose.ui.test.assertIsNotEnabled
-import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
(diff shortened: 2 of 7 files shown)


rinat-enikeev:
@lead-onym make a build

PR #94, opened 2026-05-05
ux: restore identity from 12-word recovery phrase (closes #86)

### LINKED ISSUE #86: No way to restore identity using 12-word recovery phrase
## Summary

There is currently no way to restore an existing identity using a 12-word recovery phrase (mnemonic). When a user has a backup phrase from a previously created identity, they cannot import or recover that identity into the app.

The **Identities** section has no restore or import option. The **Privacy & Encryption** section contains a "Your Keys" block with the following buttons:
- "Recovery Phrase"
- "Nostr Identifier"
- "Group BLS Key"
- "Stellar Key"

However, **none of these buttons are clickable / tappable**. They appear as static text or disabled UI elements. As a result, the user cannot initiate any recovery procedure.

---

## Steps to Reproduce

2. Navigate to **Settings** → **Privacy & Encryption** (or similar section).
3. Locate the **"Your Keys"** block.
4. Observe the four buttons: "Recovery Phrase", "Nostr Identifier", "Group BLS Key", "Stellar Key".
5. Tap on any of these buttons.
6. Observe that **nothing happens** – the buttons are not responsive.
7. Navigate to **Settings** → **Identities**.
8. Look for any "Restore", "Import", or "Recover" option.
9. Observe that no such option exists.

---

## Expected Behavior

- The user should be able to **restore an existing identity** using a 12-word recovery phrase.
- A clear **"Restore Identity"** or **"Import from Recovery Phrase"** option should be present, ideally in the Identities section.
- The "Recovery Phrase" button (or a similar UI element) should be **clickable** and should open either:
  - An input dialog to enter the 12-word phrase, or
  - A screen showing the current identity's recovery phrase (if already logged in).
- Recovery should create or switch to the restored identity within the app.

---

## Actual Behavior

- No restore / import functionality exists for 12-word mnemonics.
- The "Recovery Phrase" button and other key-related buttons are **non-clickable** (disabled or decorative).
- Users who have a backup phrase cannot recover their identity after reinstalling the app or moving to a new device.

---

ux-onym:
## Summary

Closes #86. Users can now restore an existing identity from its 12-word BIP39 recovery phrase.

**UX decision.** Restore lives in **Settings → Identities** as a sibling card to "Add Identity" (rather than wedged into Privacy & Encryption, which is documented as a read-only explainer of the cryptographic primitives — not an actions surface). Single-tap "create fresh" stays unchanged; restore gets its own clearly-labeled top-level affordance — matches the wallet/keyring convention of separate "Create new" / "Import existing" entries.

**Flow:**
1. Tap "Restore from recovery phrase" → new `RestoreIdentityScreen`.
2. Paste/type 12 (or 24) words + optional alias. Live BIP39-checksum validation gates the Restore button and drives an inline hint.
3. Submit calls the existing non-destructive `IdentityRepository.add(name, restoreMnemonic = ...)`. The restored identity lands alongside any existing ones and becomes the active selection.

All cryptographic primitives, storage, and the `add(restoreMnemonic = …)` API already existed — this PR is wiring + UI.

## Notes

- Strings localized for **en** and **ru** (matches commit fd344d7's parity contract; `StringsCompletenessTest` passes).
- No biometric gate on restore — wrong phrase yields a different identity, the user notices on first chat. Mirrors how every other BIP39 wallet treats restore.
- The Privacy & Encryption "Your Keys" block stays read-only by design (out of scope for this issue).

## Test plan

- [ ] Settings → Identities shows two action cards: "Add Identity" (blue +) and "Restore from recovery phrase" (green Restore icon).
- [ ] Tapping Restore opens the screen with a multi-line phrase input.
- [ ] Typing a partial / invalid phrase shows the red "Invalid phrase" hint and disables the Restore button.
- [ ] Pasting a valid 12-word phrase flips the hint green and enables Restore.
- [ ] Tapping Restore creates a new identity, switches to it, and pops back to Identities.
- [ ] Restoring with the same phrase that's already on the device produces the same key material (so the BLS public key matches the original identity's).
- [ ] System back from RestoreIdentityScreen returns to Identities without committing anything.
- [ ] Russian locale: every new string is translated; layout doesn't truncate.

### DIFF
--- a/app/src/main/kotlin/chat/onym/android/AppDependencies.kt
+++ b/app/src/main/kotlin/chat/onym/android/AppDependencies.kt
@@ -47,6 +47,8 @@ class AppDependencies(
     val makeChatsViewModel: () -> ChatsViewModel,
     /** Settings → Identities — multi-identity management (PR-5). */
     val makeIdentitiesViewModel: () -> chat.onym.android.identity.IdentitiesViewModel,
+    /** Settings → Identities → Restore from recovery phrase. */
+    val makeRestoreIdentityViewModel: () -> chat.onym.android.identity.RestoreIdentityViewModel,
     /** Post-create deeplink invite share (deeplink-invite PR-5). */
     val makeShareInviteViewModel: () -> chat.onym.android.group.ShareInviteViewModel,
     /** Joiner-side post-deeplink-tap surface (deeplink-invite PR-7).
--- a/app/src/main/kotlin/chat/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/chat/onym/android/OnymApplication.kt
@@ -539,6 +539,9 @@ class OnymApplication : Application() {
             makeIdentitiesViewModel = {
                 chat.onym.android.identity.IdentitiesViewModel(identity = identityRepository)
             },
+            makeRestoreIdentityViewModel = {
+                chat.onym.android.identity.RestoreIdentityViewModel(identity = identityRepository)
+            },
             makeShareInviteViewModel = {
                 chat.onym.android.group.ShareInviteViewModel(
                     identity = identityRepository,
--- a/app/src/main/kotlin/chat/onym/android/RootScreen.kt
+++ b/app/src/main/kotlin/chat/onym/android/RootScreen.kt
@@ -56,6 +56,7 @@ import chat.onym.android.settings.IdentityDetailScreen
 import chat.onym.android.settings.PrivacyEncryptionScreen
 import chat.onym.android.settings.RelayerSettingsScreen
 import chat.onym.android.settings.RelayerSettingsViewModel
+import chat.onym.android.settings.RestoreIdentityScreen
 import chat.onym.android.settings.RunRelayerScreen
 import chat.onym.android.settings.SettingsScreen
 import chat.onym.android.settings.UseExistingContractScreen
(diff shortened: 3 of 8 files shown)


dev-onym (commented):
COMMENT
- `Bip39.entropyFromMnemonic` splits only on `" "` (Bip39.kt:104), so a phrase pasted with newlines/tabs (e.g. one word per line — common from a written-down backup or text-file copy) is flagged Invalid. Suggest normalizing whitespace (`split("\\s+".toRegex())`) inside the BIP39 helper or in `RestoreIdentityViewModel.kt:46-58/68`.
- Error snackbar reuses `identities_error_add` ("Couldn't add identity: %s") on the Restore screen at `RestoreIdentityScreen.kt:80,84` — wording mismatches the action. Add a dedicated `restore_identity_error` string for both `values/strings.xml` and `values-ru/strings.xml`.
- No unit tests for `RestoreIdentityViewModel` (no hits in `app/src/test` or `app/src/androidTest`). For a key-restore flow with a non-trivial Empty/Invalid/Valid state machine and one-shot Done semantics, a small JVM test of `validation`, `submit` happy/error path, and `acknowledgeDone` would be cheap insurance.
- Phrase field sets `imeAction = ImeAction.Next` (`RestoreIdentityScreen.kt:159`) but no `keyboardActions` is wired, so Next does nothing — either drop the hint or add a `FocusRequester` for the name field.
- Restoring a phrase already on the device produces a second entry with identical keys but a fresh `IdentityId` (`IdentityRepository.kt:247-250`); the test-plan item only verifies key parity, not that the list shows a confusing duplicate. Pre-existing `add()` behavior, but this PR is the first UI that surfaces it — worth a follow-up to dedupe by public key, or at least a banner.
- Minor: validation-hint slot uses `heightIn(min = 18.dp)` (`RestoreIdentityScreen.kt:182`), but the Russian "Неверная фраза…" can wrap on narrow screens and still cause a small jump; consider `min = 36.dp` or two-line reservation.

rinat-enikeev:
@lead-onym file an issue to onym-ios repo with same ux improvement so that iOS is aligned with android

rinat-enikeev:
@lead-onym make build

rinat-enikeev:
@lead-onym make a build

rinat-enikeev:
fixed in other PR

PR #97, opened 2026-05-05
feat(group): add MemberAnnouncementPayload wire format

rinat-enikeev:
## Summary

- New `@Serializable data class MemberAnnouncementPayload` plus nested `AnnouncedMember`. Snake_case JSON keys + base64 ByteArrays so it's byte-equivalent with onym-ios.
- `commitment` + `epoch` are optional today; PRs 88 / 89 fill them in for the on-chain anchor flow. Pre-PR-88 senders ship without them and receivers stay best-effort until PR 89.
- `init` validation throws on malformed byte lengths.

Stacked on `group/approver-ui` (PR #96). Mirrors onym-ios PR #77.

## Test plan
- [ ] `./gradlew :app:testDebugUnitTest` green
- [ ] `MemberAnnouncementPayloadTest` covers decode-with/without optionals, encode omits nulls, byte-length validation throws, unknown fields are tolerated.

### DIFF
--- a/app/src/main/kotlin/chat/onym/android/group/MemberAnnouncementPayload.kt
+++ b/app/src/main/kotlin/chat/onym/android/group/MemberAnnouncementPayload.kt
@@ -1,0 +1,155 @@
+package chat.onym.android.group
+
+import chat.onym.android.identity.Base64ByteArraySerializer
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Plaintext payload that the admin seals (via
+ * `IdentityRepository.sealInvitation`, X25519 + AES-GCM, signed by
+ * the admin's Ed25519 stellar key) and ships to every existing
+ * member's inbox after they Approve a join request. Tells receivers
+ * "this person just joined the group — append them to your local
+ * roster".
+ *
+ * Sits alongside [GroupInvitationPayload]:
+ *   - [GroupInvitationPayload] is the joiner's first taste of a
+ *     group — full state needed to render messages.
+ *   - [MemberAnnouncementPayload] is incremental — existing members
+ *     already know the group, they just need to learn about one new
+ *     entry in the roster.
+ *
+ * ## Trust
+ *
+ * `newMember.alias` and [adminAlias] are self-asserted. Receivers
+ * that care about provenance should display the BLS-pubkey
+ * fingerprint alongside (matches the inviter-approval guidance on
+ * [JoinRequestPayload]). The OUTER `SealedEnvelope` carries the
+ * admin's Ed25519 signature over the ephemeral key — receivers MUST
+ * cross-check `senderEd25519PublicKey` against the group's stored
+ * `adminEd25519PubkeyHex` (PR 84) before mutating local state. That
+ * signature check lives in the dispatcher, not here — this type is
+ * a pure value carrier.
+ *
+ * ## Versioning
+ *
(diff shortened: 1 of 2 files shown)


dev-onym:
Quick read — shape and tests mirror onym-ios PR #77 cleanly. A few things worth nailing down before the seal site lands in PR 88:

- **Encoder config is load-bearing for cross-platform parity, and the test masks it.** The `encode_omitsNullOptionalsWhenNotSet` test only passes because its local `Json` is `encodeDefaults = false`. The rest of the group package (`CreateGroupInteractor`, `JoinRequestSender`, `JoinRequestApprover`, `RoomGroupStore`) all use `encodeDefaults = true`. If `MemberAnnouncementPayload` gets serialized through one of those, you'll emit `"commitment": null` / `"epoch": null` — and Swift's synthesized Codable on iOS uses `encodeIfPresent` and omits, so the wire bytes won't match. `IntroCapability` already chose `encodeDefaults = false` for the same reason. Either pin a companion `Json` on this type, or have PR 88's sender explicitly use `encodeDefaults = false; ignoreUnknownKeys = true`.
- **`version` is purely declarative.** No `require(version == 1)` (matches `GroupInvitationPayload`'s posture, but worth a deliberate decision). Per the KDoc the forward-compat path is "nullable defaults," which means a v=2 announcement that only adds optional fields decodes fine here — but the test suite doesn't pin that contract. A `decode_v2WithUnknownOptional` test would lock it in.
- **`epoch: ULong` over JSON** is fine in practice (Tyranny epochs won't approach 2^53) but Swift `UInt64` ↔ Kotlin `ULong` byte-equivalence isn't free past that range. One-line KDoc would save the next person from widening it.
- **Golden vector** — none of the round-trip tests anchor against a fixed JSON byte string emitted by the iOS counterpart. Even one shared canonical-bytes vector (same one in onym-ios PR #77) would catch any future drift before it ships rather than at first cross-platform send.

For a structured pass, kick off a review via the request-review flow. For a specific change, drop `@claude <instruction>` on the relevant line.

dev-onym:
Thanks for the ping — no new instruction in the PR body, so just a status pass against what's there now (`a01b774`):

- **Code + tests look unchanged from my earlier comment.** The four follow-ups from [#issuecomment-4383001587](https://github.com/onymchat/onym-android/pull/97#issuecomment-4383001587) are all still open: encoder-config parity (the `encodeDefaults = false` local `Json` masks the rest of the group package's `encodeDefaults = true`), no `version` guard, no `epoch: ULong` cross-platform note, and no shared golden vector against onym-ios PR #77.
- **Of those, only the encoder-config one is a true blocker before PR 88's seal site** — it'll silently emit `"commitment": null` / `"epoch": null` through `CreateGroupInteractor`/`JoinRequestApprover`'s shared `Json` and break wire-byte parity with Swift's `encodeIfPresent`. Easiest fix is a `companion object { val Json = Json { encodeDefaults = false; ignoreUnknownKeys = true } }` on the type, mirroring `IntroCapability`.
- The test plan items themselves are covered by what's already in `MemberAnnouncementPayloadTest` — decode-with/without optionals (✔), encode omits nulls (✔ but only under the local Json), byte-length validation throws (✔ via `init`), unknown fields tolerated (✔ via `ignoresUnknownFields`).

For a structured pass over the file, kick off the request-review flow. For a specific change (e.g. "pin a companion `Json` on this type"), drop `@claude <instruction>` on the relevant line.

Issue #83, opened 2026-05-03
Allow duplicate Identity names; suggests custom naming on creation


alexpovstin:
## Summary

When creating a new "Identity" (profile), the app automatically assigns a default name in the format `Identity N` (where N is a sequential number). The app does **not** check whether an identity with the same name already exists. Multiple identities can therefore have identical names (e.g., two or more identities both named `Identity 1`). This leads to confusion, especially when managing multiple profiles or switching between them.

The tester suggests that a better user experience would be to:
- **Prompt the user to enter a custom name** during identity creation
---

## Steps to Reproduce

1. Open the app on an Android smartphone.
2. Go to **Settings** → **Identities** (or profile management section).
3. Tap **Create New Identity**.
4. The app instantly creates a new identity with default name `Identity

alexpovstin:
@lead-onym What suggestions do you have?

lead-onym:
**Routed to @dev-onym.**

_

@dev-onym will respond shortly. Cross-mention me (`@lead-onym`) on this thread to override and re-route.

PR #103, opened 2026-05-05
feat(inbox): joiner-side group materialization from invitation

rinat-enikeev:
## Summary

- Second fast-path on `IncomingMessageDispatcher`: a decoded `GroupInvitationPayload` materializes a `ChatGroup` row under the receiver identity. The chat now appears in the chats list without going through any "Accept invitation?" UI.
- Directory is the union of wire-shipped `member_profiles` (PR 82) + the receiver's own `MemberProfile` looked up from `IdentityRepository.identities`. Self-last so the sender can't assert our alias.
- Bad `tier_raw` / `group_type_raw` drops the message rather than materializing a partial group. Idempotent on `groupId`.

PR 84 will populate `adminEd25519PubkeyHex` from the envelope's sender pubkey at this point. PRs 89 / 91 will gate Tyranny invitations against the on-chain commitment.

Stacked on `group/invitation-member-profiles` (PR #102). Mirrors onym-ios PR #83.

## Test plan
- [ ] `./gradlew :app:testDebugUnitTest` green
- [ ] `IncomingMessageDispatcherTest.invitation_materializesGroupAndAddsSelfProfile` — new group lands; directory has wire entries + self
- [ ] `IncomingMessageDispatcherTest.invitation_redeliveryIsIdempotent` — re-delivery doesn't mint duplicates
- [ ] Existing announcement / fall-through tests still pass

### DIFF
--- a/app/src/main/kotlin/chat/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/chat/onym/android/OnymApplication.kt
@@ -325,6 +325,7 @@ class OnymApplication : Application() {
             envelopeDecrypter = identityRepository,
             groupRepository = groupRepository,
             invitationsRepository = invitationsRepository,
+            identitiesFlow = identityRepository.identities,
         )
         val invitationsInteractor = chat.onym.android.inbox.IncomingInvitationsInteractor(
             inboxTransport = inboxTransport,
--- a/app/src/main/kotlin/chat/onym/android/inbox/IncomingMessageDispatcher.kt
+++ b/app/src/main/kotlin/chat/onym/android/inbox/IncomingMessageDispatcher.kt
@@ -1,13 +1,17 @@
 package chat.onym.android.inbox
 
+import chat.onym.android.chain.SepGroupType
+import chat.onym.android.chain.SepTier
 import chat.onym.android.group.ChatGroup
+import chat.onym.android.group.GroupInvitationPayload
 import chat.onym.android.group.GroupRepository
 import chat.onym.android.group.MemberAnnouncementPayload
 import chat.onym.android.group.MemberProfile
-import chat.onym.android.identity.ActiveIdentityProvider
 import chat.onym.android.identity.DecryptedEnvelope
 import chat.onym.android.identity.IdentityId
+import chat.onym.android.identity.IdentitySummary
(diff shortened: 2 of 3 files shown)


dev-onym (commented):
APPROVE — solid second fast-path with proper idempotency, ordered decode, and self-last directory merge. Test coverage matches the PR description.

- `IncomingMessageDispatcher.kt:130-132` self-last merge correctly overwrites any sender-asserted entry under our own BLS key. Matches the docstring claim.
- `IncomingMessageDispatcher.kt:90-94` invitation decode is attempted *after* announcement; required-field sets are disjoint enough (invitation needs `group_secret`/`members`/`salt`/`tier_raw`/`group_type_raw`; announcement needs `new_member`/`admin_alias`) that mis-routing is not a practical risk with `ignoreUnknownKeys = true`.
- `IncomingMessageDispatcher.kt:139` uses `System.currentTimeMillis()` for `createdAtMillis` rather than the envelope's `receivedAt`. Minor — `receivedAt` is already a parameter, would give a slightly more truthful local creation time on backfill/re-delivery.
- `IncomingMessageDispatcher.kt:53` `identitiesFlow: StateFlow<List<IdentitySummary>>? = null` — the nullable-with-default is purely for test ergonomics; only one test actually exercises the self-backfill. Consider dropping the default so the production wiring is type-enforced and missed test setups become compile errors instead of silent skips. Not blocking.
- `IncomingMessageDispatcher.kt:121` `senderEd25519PublicKey` is plumbed but suppressed-unused; intentional per PR-84 deferral. Worth tracking — until PR 84 lands, materialization trusts `adminPubkeyHex` from the payload without cross-checking against the envelope sender. Per PR description this is by design and gated chain-side later.
- `IncomingMessageDispatcherTest.kt:265-267` redelivery test asserts `groups.size == 1` against a filtered snapshot — good. Combined with `GroupRepository.insert → GroupStore.insertOrUpdate`, idempotency is properly verified.
- `IncomingMessageDispatcher.kt:168-175` `catch (_: Throwable)` is broader than needed (the serializer raises `SerializationException`); not harmful since this is not a suspend boundary, but `catch (_: SerializationException)` would document intent better. Mirrors the existing `tryDecodeAnnouncement` style, so leaving it consistent is also fine.

Issue #117, opened 2026-05-08
Contract not allowlisted for tyranny on testnet when creating group


alexpovstin:
## Summary

When creating a **tyranny group** (named "smoke"), the process fails at the final stage ("Anchor in stellar") with an HTTP 400 error indicating the contract ID is not allowlisted for tyranny on testnet.

The error occurs consistently across two different Android smartphones (Samsung Galaxy A50 and Tecno KL6), with both empty groups and groups that already have one participant added.

---

## Error Message (exact)

Couldn't reach the relayer: contract call returned HTTP 400: {"accepted":false,"message":"contractID CAFX4A2KLOK7RE5QSPTDQ3UBCWOUSBPPFC2PU7M6ZP53GPGRCL67HNR3 is not allowlisted for tyranny on testnet"}

---

## Steps to Reproduce

1. Open the app on an Android smartphone.
2. Start creating a new **tyranny group** (e.g., named "smoke").
3. Enter the group name.
4. Create an empty group (no participants added, or with one participant).
5. Proceed through the creation stages:
   - **Stage 1:** "Setting up encrypted group" / "Creating keys" → Success
   - **Stage 2:** "Setting up admin keys" / "You will be the only admin" → Success after ~30 seconds
   - **Stage 3:** "Sending invitation" → Success
   - **Stage 4:** "Anchor in stellar" → **Error**
6. Observe the HTTP 400 error as shown above.

alexpovstin:
onym v.0.0.16-d8d71

alexpovstin:
@lead-onym identify the cause of the described error

lead-onym:
**Routed to @dev-onym.**

_

@dev-onym will respond shortly. Cross-mention me (`@lead-onym`) on this thread to override and re-route.

dev-onym:
## Diagnosis

The Android app is behaving correctly; the rejection is on the **relayer** side.

**Cause:** the production relayer's allowlist (`RELAYER_CONTRACT_ALLOWLIST`) hasn't been refreshed to include the contract addresses from `onym-contracts` release **v0.0.11** (published 2026-05-07).

### Evidence

1. The contract ID in the error — `CAFX4A2KLOK7RE5QSPTDQ3UBCWOUSBPPFC2PU7M6ZP53GPGRCL67HNR3` — matches the `(testnet, tyranny)` entry in the latest manifest:
   ```
[...]

PR #119, opened 2026-05-08
fix: actionable error when relayer rejects un-allowlisted contract (#117)

dev-onym:
## Summary

- Detect the relayer's "contractID … is not allowlisted for <type> on <network>" rejection (issue #117) and surface a distinct, actionable error instead of the raw transport text.
- Add `CreateGroupError.AnchorContractNotAllowlisted` carrying `(contractId, governance, network, release)` so the UI can render the precise remediation: "Open Settings → Anchors and pick an earlier release."
- Cover both the live HTTP-400 path and the defensive HTTP-200 / `accepted:false` fallback.

## Why this PR

When a new `onym-contracts` release lands before the relayer's allowlist is rolled (the actual root cause of #117 — v0.0.11 contracts not yet allowlisted on `relayer.onym.chat`), every Tyranny / Anarchy / 1-on-1 create on testnet fails. The relayer is the right side to fix it operationally, but the Android client should still fail informatively: until the relayer is updated, the user has a one-tap escape via Settings → Anchors → pick an earlier release. The current error swallows that affordance under the generic `Couldn't reach the relayer: contract call returned HTTP 400: {…}` banner.

## Test plan

- [x] `:app:compileDebugSources` clean
- [x] New unit tests in `CreateGroupInteractorTest`:
  - `create_anchorRejected_notAllowlistedBody_throwsContractNotAllowlisted` — HTTP 400 + `is not allowlisted` body
  - `create_anchorRejectedWith200_notAllowlistedMessage_throwsContractNotAllowlisted` — defensive HTTP-200 fallback
- [x] Existing `create_anchorTransportError_throws` (HTTP 502) and `create_anchorRejected_throws` (rejected with unrelated message) still hit their original branches.
- [ ] Manual: with v0.0.11 selected on testnet, attempt to create a Tyranny group → confirm new error banner; tap Settings → Anchors → pick v0.0.5 → retry succeeds.

### DIFF
--- a/app/src/androidTest/kotlin/chat/onym/android/group/CreateGroupInteractorTest.kt
+++ b/app/src/androidTest/kotlin/chat/onym/android/group/CreateGroupInteractorTest.kt
@@ -299,6 +299,56 @@ class CreateGroupInteractorTest {
     }
 
     @Test
+    fun create_anchorRejected_notAllowlistedBody_throwsContractNotAllowlisted() = runBlocking {
+        // The relayer rejects an un-allowlisted contract with HTTP 400
+        // and a body of the shape:
+        //   {"accepted":false,"message":"contractID C… is not allowlisted for tyranny on testnet"}
+        // Issue #117 — the user should see a distinct, actionable error
+        // rather than the raw transport message.
+        contractTransport.setBehavior(
+            ConfigurableContractTransport.Behavior.Throws(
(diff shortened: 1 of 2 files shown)


alexpovstin:
@lead-onym build apk

alexpovstin:
the group creation error has been fixed

PR #120, opened 2026-05-09
chain(anarchy): publish "not tracked" sentinel for member_count at create

rinat-enikeev:
## Summary

- `sep-anarchy`'s `create_group` accepts `member_count: 0` as the documented "not tracked" sentinel. The contract treats `member_count` as informational, never mutates it, and the test-vector pin (`anarchy/v1 sep-xxxx`) reserves `0` for the "not tracked" case.
- Production was passing `members.size` — always `1` at create, since the founding roster is creator-only. That published a small but permanent piece of group structure (`tier=Small, member_count=1`) into the on-chain `GroupCreated` event for every Anarchy group ever created via the Android app.
- Switch to the sentinel. Chain observers now learn only the size class (`tier`), not the exact roster size at create. Accurate counts still live in the local model and the UI — only the wire value sent to the relayer changes.

This is the Android half of a coordinated change with `onymchat/onym-ios` (matching branch / PR title). Web copy on `/humans` and `/threat-model` will be tightened in a follow-up to take advantage of the narrower on-chain footprint.

## Test plan

- [ ] `./gradlew test` — `CreateGroupInteractorTest` and `SepContractTypesTest` / `SepContractClientTest` should pass. The encoder tests are unchanged (they pin `1` to verify u32 serialization works); only the production callsite changed.
- [ ] `./gradlew connectedAndroidTest` — `CreateGroupAnarchyE2ETest.create_zeroInvitees_anchorsOnTestnet` should still anchor and round-trip. The on-chain assertions only check commitment + epoch, so the `member_count` change does not affect them. The doc-comment is refreshed.
- [ ] Manual: create an Anarchy group on testnet and inspect the `GroupCreated` event payload — `member_count` should be `0`.
- [ ] Confirm the local UI still shows the correct count (`group.members.size`); the on-chain value isn't read back into the UI.

### DIFF
--- a/app/src/androidTest/kotlin/chat/onym/android/integration/CreateGroupAnarchyE2ETest.kt
+++ b/app/src/androidTest/kotlin/chat/onym/android/integration/CreateGroupAnarchyE2ETest.kt
@@ -130,8 +130,9 @@ class CreateGroupAnarchyE2ETest {
     /**
      * Happy path with zero invitees. Verifies the group anchors on
      * chain and `get_commitment` round-trips the same commitment
-     * back. Anarchy tolerates a creator-only roster (the contract's
-     * `member_count` is informational; passing `1` is fine).
+     * back. Anarchy tolerates a creator-only roster; production sends
+     * `member_count = 0` (the contract's documented "not tracked"
+     * sentinel) so the chain only learns the tier.
      */
     @Test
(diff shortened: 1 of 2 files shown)


rinat-enikeev:
@lead-onym create an issue to onym-contracts repo to remove members_count from anarchy contract

PR #127, opened 2026-05-10
feat(release-issue): auto-bump patch tag when Release tag is blank

rinat-enikeev:
## Summary

File a release issue without filling in the **Release tag** field and the `release-from-issue` workflow now resolves the next patch version itself — `v0.0.38 → v0.0.39` against current `main` — and dispatches Release with that. Mirrors onymchat/onym-ios#123.

### Issue template
- `Release tag` is no longer required. New help text + placeholder explain the auto-bump behaviour.

### `scripts/release_issue.py`
- **`parse-tag`** returns `""` (exit 0) for both blank fields and GitHub's `_No response_` placeholder. Malformed values still abort with the existing error message — only the missing-value path changed.
- **`hydrate`** gains an optional `--tag vX.Y.Z` flag that rewrites the `### Release tag` field's value via a new `_rewrite_tag()` helper. Idempotent for the explicit case; meaningful for the auto-bumped case so the issue body itself shows what got dispatched.

### `release-from-issue.yml`
- `actions/checkout@v4` now uses `fetch-tags: true` so the runner sees `v*` tags (default depth-1 fetch wouldn't).
- **Resolve release tag** step (renamed from "Parse + validate"):
  - On empty `parse-tag` output: picks latest `v[0-9]+\.[0-9]+\.[0-9]+` tag, bumps patch, falls back to `v0.0.1` on a tagless repo. Emits an `auto_bumped` step output.
  - Refuses to dispatch on a tag that already exists — protects a hand-typed value that collides with a previous release. The auto-bump path is collision-free by construction.
- **Body-rewrite step** passes `--tag $TAG` to `hydrate`, then comments on the issue when auto-bumped so the maintainer can correct it (close + reopen with a different value) if needed.

## Test plan

Verified locally:
- [x] `parse-tag` returns the typed tag, `""` for `_No response_`, and rejects `0.0.45` (missing `v`) with the existing error.
- [x] `hydrate --tag v0.0.39` rewrites `### Release tag` and fills the UI_TESTS section in a single pass.
- [x] Auto-bump arithmetic against `git tag --sort=-v:refname` on current `main` produces `v0.0.38 → v0.0.39`.

Reviewer checklist:
- [ ] File a release issue with the Release tag field blank — confirm the workflow rewrites the body to show the auto-bumped tag, leaves the explanation comment, and dispatches Release for the right tag.
- [ ] File a release issue with a colliding tag (e.g. `v0.0.38`) — confirm the workflow comments + fails before dispatch.
- [ ] File a release issue with a properly-typed new tag — confirm behaviour matches the previous workflow.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release.yml
+++ b/.github/ISSUE_TEMPLATE/release.yml
@@ -17,10 +17,14 @@ body:
     id: tag
     attributes:
       label: Release tag
-      description: Must match `vX.Y.Z`. The Release workflow validates this and aborts if the tag already exists.
-      placeholder: v0.0.8
+      description: |
+        `vX.Y.Z`. Leave blank to auto-bump the patch version from the latest
+        existing `v*` tag — the `release-from-issue` workflow rewrites this
+        field in the issue body with the resolved tag before dispatching
+        Release. Submitting an explicit tag that already exists fails.
+      placeholder: v0.0.8 (or leave blank to auto-bump)
     validations:
-      required: true
+      required: false
 
   - type: textarea
     id: notes
--- a/.github/workflows/release-from-issue.yml
+++ b/.github/workflows/release-from-issue.yml
@@ -38,6 +38,11 @@ jobs:
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
+        with:
+          # Tag history is needed so the auto-bump branch can resolve
+          # the latest patch version when the issue's Tag field is left
+          # blank. `fetch-tags` is lighter than `fetch-depth: 0`.
+          fetch-tags: true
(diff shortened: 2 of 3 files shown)


dev-onym:
Thanks for the mention — I don't see a specific question or instruction in the PR body, just the Claude Code attribution line, so I don't have a clear ask to act on here.

If you'd like me to take a look:
- For a structured review of the diff, please use the request-review flow on the PR.
- For a targeted code change, drop a line-level comment with `@claude <instruction>` on the specific file/line you want adjusted (e.g. on `scripts/release_issue.py` or `.github/workflows/release-from-issue.yml`).

rinat-enikeev:
@dev-onym fix invalid workflow file in this PR, see https://github.com/onymchat/onym-android/actions/runs/25623239049

PR #131, opened 2026-05-10
feat(release): publish signed APK to onym.app/qa/ alongside GitHub Release

### LINKED ISSUE #130: Release: vX.Y.Z
### Release tag

v0.0.42

### Release notes (optional)

_No response_

## UI tests

- [x] chat.onym.android.chain.GroupProofGeneratorFfiTest
- [x] chat.onym.android.group.CreateGroupInteractorTest
- [x] chat.onym.android.group.GroupCommitmentBuilderFfiTest
- [x] chat.onym.android.identity.IdentityRepositoryInvitationDecryptTest
- [x] chat.onym.android.identity.IdentityRepositorySealInvitationTest
- [x] chat.onym.android.identity.IdentityRepositoryTest
- [x] chat.onym.android.integration.CreateGroupAnarchyE2ETest
- [x] chat.onym.android.integration.CreateGroupOneOnOneE2ETest
- [x] chat.onym.android.integration.CreateGroupTyrannyE2ETest
- [x] chat.onym.android.recovery.RecoveryPhraseBackupScreenTest
- [x] chat.onym.android.recovery.RecoveryPhraseBackupViewModelTest
- [x] chat.onym.android.transport.nostr.OnymNostrSignerTest
- [x] chat.onym.android.uitests.AnchorsUITest
- [x] chat.onym.android.uitests.IdentitiesUITest
- [x] chat.onym.android.uitests.RelayerSettingsUITest

### Install + onboarding
- [ ] Sideload the signed APK on a real device (Android 14+ recommended)
- [ ] First-launch onboarding flow renders without crash
- [ ] Bootstrap a new identity, then restart the app and confirm it persists
- [ ] Restore identity from a 12-word mnemonic, confirm pubkey matches

### Create Group — Tyranny
- [ ] Create an empty Tyranny group against testnet, confirm it lands on the Stellar contract
- [ ] Create a Tyranny group with 1 invitee, confirm invitee receives sealed envelope
- [ ] Open Settings → toggle Mainnet, confirm next create resolves the mainnet contract binding

### Create Group — 1-on-1
- [ ] Pick the 1-on-1 governance card, paste an invitee inbox key, confirm "Start 1-on-1" CTA enables only at exactly 1 invitee
- [ ] Create the 1-on-1 group, confirm anchor on testnet
- [ ] Confirm invitee receives sealed envelope carrying the ephemeral `sk_1`

### Settings + relayer + anchors
- [ ] Settings → Relayer Settings → swipe-delete an endpoint, confirm primary marker clears if deleted
- [ ] Settings → Anchors → drill into Testnet → Tyranny, pick an older release, confirm persistence
- [ ] Add a custom relayer URL, confirm it lands as `custom` network endpoint

### Theming + chrome
- [ ] Toggle system dark mode, every Create Group screen renders correctly in both themes
- [ ] Bottom-tab Chats / Settings / Search switching works without flicker
- [ ] Recovery-phrase backup flow (Settings → Backup) walks Intro → Reveal → Verify → Done

### Production sanity
- [ ] APK installs without "package conflicts with existing" prompt over the prior release
- [ ] Network Settings shows the deployed `relayer.onym.chat` endpoint by default after auto-populate
- [ ] AltStore source updates with the new IPA URL once notarization mirrors land (separate workflow — leave unchecked if N/A)

rinat-enikeev:
Closes #130.

## Why this release didn't show up on onym.app/qa/

A successful tag-gated `Release` (`v0.0.42` from #130) attached the signed APK to a GitHub Release and stopped — there is no step in `release.yml` that stages a manifest, merges into `builds.json`, or rsyncs to onym.app. The QA dashboard is populated only by the manually-dispatched `qa-build.yml`, so testers couldn't tap "Install" for an actual release. iOS landed the equivalent fix in onymchat/onym-ios#126 (already merged); this PR mirrors it.

## What changed

In `release.yml`'s `build` job, immediately after `softprops/action-gh-release@…` (the GitHub Release upload):

- **Stage signed APK** under `dist/qa/android/<tag>/onym.apk` (same file path the dashboard expects).
- **Build dashboard entry** (`kind: tag`, `branch: null`, `description: "Release vX.Y.Z"`).
- **Merge into `builds.json`** — fetch the live one, prepend our entry, `unique_by(platform+@+version)` so a re-run for the same tag cleanly overwrites, sort newest-first, cap 50.
- **rsync** via `onymchat/onym-website/.github/actions/deploy@main` with `delete_stale: false` so iOS uploads + earlier Android builds aren't wiped.
- **Step summary table** with install + dashboard URLs.
- **Issue comment** with the install URL when dispatched from a release issue (best-effort — a `gh` failure leaves a `::warning::` but never fails the release run).

### Concurrency

Added a job-level `qa-build-android` group on the release `build` job — same group `qa-build.yml` uses at workflow level. Tag-gated releases and manual QA dispatches now serialize on `builds.json` writes; lint + tests still run in parallel.

### Why inline rather than a composite action

Same trade-off iOS made: the QA-publish chain is now duplicated between `qa-build.yml` and `release.yml`. Refactoring into `.github/actions/qa-publish-android/action.yml` would deduplicate but adds an indirection layer for two callers. Kept inline so both flows stay readable side-by-side; easy refactor opportunity if a third caller ever appears.

## Test plan

- [x] `python3 -c 'import yaml; yaml.safe_load(open(...))'` parses the workflow cleanly.
- [ ] Reviewer files a fresh release issue (auto-bump path) and confirms:
  - The APK lands as a GitHub Release asset (existing behavior).
  - `https://onym.app/qa/android/<tag>/onym.apk` is reachable + installs on a real device.
  - `https://onym.app/qa/?platform=android` shows the new entry with `kind: tag`.
  - The release issue receives a comment with the install URL.
- [ ] Reviewer dispatches a manual `QA Build` while a release is publishing — confirms the second waits on the `qa-build-android` lock instead of racing on `builds.json`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -253,6 +253,13 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     needs: [lint, test, android-test, create-release]
+    # Share the lock with qa-build.yml so a tag-gated release and a
+    # manual ad-hoc QA dispatch don't race on /qa/builds.json. Only
+    # the build job (where the QA publish runs) takes the lock; lint
+    # and tests still run in parallel.
+    concurrency:
+      group: qa-build-android
+      cancel-in-progress: false
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
         with:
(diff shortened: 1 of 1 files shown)


dev-onym:
Skimmed the diff and cross-checked against `qa-build.yml`. A few observations:

**Looks right**
- The job-level `concurrency: qa-build-android` correctly matches `qa-build.yml`'s workflow-level group (`qa-build.yml:54`), so a release `build` job and a manual QA dispatch will serialize on `builds.json` writes while the release's `lint` / `test` / `android-test` gates still run in parallel.
- Hardcoding `branch: null` and dropping the `(if $tag != "" then $tag else null end)` conditional is consistent with `release.yml`'s `inputs.tag` being `required: true` — there is no branch-build path to defend against here.
- The `unique_by(.platform + "@" + .version)` + lead-with-our-entry pattern means a re-run for the same tag cleanly overwrites the dashboard row with this run's metadata, which matches the PR description.
- Posting straight via `gh issue comment` (rather than `qa-build.yml`'s PR-detection branch) is correct since `release-from-issue.yml` only ever plumbs an issue number, never a PR.
- `delete_stale: false` is necessary, not just defensive — `dist/qa/` only contains this run's APK + the merged `builds.json`, so deletion would wipe `qa/ios/` and prior Android tag dirs.

**Worth surfacing**
- Failure mode: if `softprops/action-gh-release` succeeds but a later QA-publish step fails (e.g. droplet rsync hiccup), the GitHub Release is *cut* but the dashboard isn't updated, and `report-to-issue` only sees the aggregate job status. Not a blocker — manual re-run resolves it because of the dedupe — but a comment on the "Comment QA install URL on the linked release issue" step pointing this out (and confirming that re-running the workflow is the recovery path) would save a future on-caller some head-scratching.
- Lock granularity: the lock currently spans the whole `build` job (~10 min of `assembleRelease` + sign + verify), not just the publish portion. A QA dispatch arriving mid-release waits for all of that, even though only the `Build dashboard entry` → `Publish` window actually contends on `builds.json`. Fine for now given dispatch volume; flagging in case it bites later.
- Test-plan item 1 (`yaml.safe_load`) is an offline parse; the cross-job concurrency group is only validated end-to-end at runtime. The reviewer's last test-plan checkbox (manual dispatch during a release) is the actual verification — keep that one as a hard gate before merging.

For a formal review pass, hit the request-review button on the PR. If you want me to revise something in-tree, leave a comment of the form `@onymchat-claude <instruction>` on the relevant file/line and I'll act on it there.

PR #141, opened 2026-05-19
feat(chats): ChatMessagePayload wire type (PR 1/3 chat stack)

lead-onym:
First PR in the Android chat stack. Adds the plaintext wire-format payload that future PRs will encrypt + transmit. **No wiring yet** — pure data type plus tests. Mirrors [onym-ios PR #147](https://github.com/onymchat/onym-ios/pull/147).

## What's in here

- `ChatMessagePayload` (`@Serializable` data class): `version`, `messageId` (UUID for dedup), `groupId` (32-byte), `senderBlsPubkeyHex` (matches `ChatGroup.memberProfiles` keying), `sentAtMillis` (Long for cross-platform clarity), `variant`.
- `ChatMessageVariant` sealed interface, governance-keyed:
  - Only `Tyranny(body: String)` ships today.
  - Discriminator on the wire is `SepGroupType.wireValue` (`"tyranny"`, `"oneonone"`, `"anarchy"`, …) — same spelling the relayer, Stellar contracts, and iOS twin already use, so future kinds slot in with no rename.
  - Unknown / known-but-unsupported kinds throw on decode (so PR A4's dispatcher will catch+drop rather than silently coerce).
- Hand-written `ChatMessageVariantSerializer` — kotlinx's polymorphic sealed-class machinery would either nest under a `type` envelope or silently route unknown kinds through a fallback, neither of which match the iOS wire shape.
- `UuidStringSerializer` — pins UUID to iOS's uppercase 36-char `UUID.uuidString` form so byte-for-byte round-trip with iOS holds.
- 9 tests covering: round-trip (ASCII / empty / unicode), wire-format pinning (snake_case keys, discriminator value `"kind":"tyranny"`, UUID uppercase, base64 `group_id`), decode failures (unknown kind, unsupported kind, missing required field, missing kind), forward-compat (extra top-level fields ignored when lenient).

## Why a governance-keyed variant in PR A1

1-on-1 will need per-message ratchet state and Anarchy will need a per-member signature. Baking the variant shape in now means turning a new governance type on for chat is just one new sealed-interface case — no schema migration.

## Divergence from iOS

- **`enum` with associated values → `sealed interface` with data-class cases.** Kotlin's idiomatic shape for the same algebraic-data-type pattern. `body` is exposed as an abstract property so future variants can share the accessor.
- **`extension ChatMessageVariant: Codable` → hand-written `KSerializer`.** Same reason iOS wrote a manual `init(from:)` — the polymorphic envelope shapes that come for free don't match the flat `{"kind":"…","body":"…"}` wire pin and would silently swallow unknown kinds.
- **`UUID` Codable default → `UuidStringSerializer`.** kotlinx-serialization has no built-in `java.util.UUID` serializer; iOS encodes UUID as uppercase via `JSONEncoder.dataEncodingStrategy` defaults, so we mirror that exactly.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest --tests 'app.onym.android.chats.ChatMessagePayloadTest'` — 9/9 pass
- [x] Full `:app:testDebugUnitTest` suite — 441/441 pass, 0 failures, 0 errors, 0 regressions

## Plan context

PR 1 of 3 in the Android chat-stack prefix that mirrors the iOS PRs 1–3 of their 10-PR chain ending in two users in a Tyranny group exchanging plain text. Next PRs:
- **PR A2** ([onym-ios #148](https://github.com/onymchat/onym-ios/pull/148)): persistence layer (`PersistedMessage` + `MessageRepository`)
- **PR A3** ([onym-ios #149](https://github.com/onymchat/onym-ios/pull/149)): plumb `sendingPubkey` through `MemberProfile` so PR A4's dispatcher can verify envelope signatures

### DIFF
--- a/app/src/main/kotlin/app/onym/android/chats/ChatMessagePayload.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatMessagePayload.kt
@@ -1,0 +1,214 @@
+package app.onym.android.chats
+
+import app.onym.android.chain.SepGroupType
+import app.onym.android.identity.Base64ByteArraySerializer
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.descriptors.buildClassSerialDescriptor
+import kotlinx.serialization.descriptors.element
+import kotlinx.serialization.encoding.CompositeDecoder
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+import kotlinx.serialization.encoding.decodeStructure
+import kotlinx.serialization.encoding.encodeStructure
+import java.util.Locale
+import java.util.UUID
+
+/**
+ * Plaintext chat-message payload that gets sealed (X25519 + AES-GCM
+ * via `IdentityRepository.sealInvitation`) and dropped on
+ * `InboxTransport.send` for each group member — same envelope path
+ * as `GroupInvitationPayload`. One envelope per recipient inbox
+ * key; the encrypted bytes carry this struct verbatim.
+ *
+ * [variant] is governance-keyed because each group flavour signs and
+ * routes messages differently down the road: Tyranny only needs the
+ * body, but a 1-on-1 dialog will carry per-message ratchet state and
+ * Anarchy will carry a per-member signature. Today only
+ * [ChatMessageVariant.Tyranny] ships — adding a new case is the
+ * entire surface-area cost of turning a new governance type on for
+ * chat.
+ *
+ * Versioned for forward compat. `version = 1` is the only shape the
+ * receiver has to handle today; later versions can add fields with
+ * non-failing `decodeIfPresent`-style decoders.
+ *
+ * Mirrors `ChatMessagePayload.swift` from onym-ios PR #147.
+ */
+@Serializable
+data class ChatMessagePayload(
+    /** Wire schema version. Bump on a breaking field change (rename,
+     *  removal, semantic shift). Adding optional fields does NOT
+     *  require a bump. */
+    val version: Int,
+    /**
+     * Stable per-message identifier used for receive-side dedup —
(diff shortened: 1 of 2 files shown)


dev-onym (commented):
**APPROVE** — clean wire-type PR with thorough tests; all 9 tests pass and integrations check out.

- `ChatMessagePayload.kt:215` — `Locale.ROOT` on UUID uppercase is correct (avoids Turkish-locale `I` bug). 
- `ChatMessagePayload.kt:160-181` — manual descriptor lists `kind`/`body` and `else -> throw` on unknown indices; with `ignoreUnknownKeys=true`, kotlinx skips unknowns before they reach the `else`, so forward-compat for *known* variants works. Worth a test for "extra field inside variant ignored when lenient" mirroring the top-level test at `ChatMessagePayloadTest.kt:198`, but not a blocker.
- `ChatMessagePayload.kt:75-94` — hand-written `equals`/`hashCode` correctly use `contentEquals`/`contentHashCode` for `groupId: ByteArray` (data class default would have been reference-eq — easy to miss).
- `ChatMessagePayload.kt:48` — doc says "32-byte" `groupId` and "96 chars" `senderBlsPubkeyHex`, but no `init {}` validation. Fine for a pure wire-type since the dispatcher in PR A4 will reject on group lookup, but worth flagging that a malformed payload from a misbehaving peer round-trips here.
- `SepContractTypes.kt:37-56` confirms `SepGroupType.wireValue` spellings (`tyranny`, `oneonone`, `anarchy`, `democracy`, `oligarchy`) match the serializer's `when` and the test's `"jellyfish"`/`"oneonone"` rejection cases — discriminator pin is honest.
- `Base64ByteArraySerializer` reused from existing group payloads (`GroupInvitationPayload`, `JoinRequestPayload`) — keeps base64 encoding consistent with iOS twin across all envelope types.
- Test coverage is genuinely thorough for a wire-type: round-trip (ASCII/empty/unicode), key pinning, discriminator pinning, UUID case pinning, base64 round-trip, 4 reject paths, forward-compat. Matches what the PR description claims.

alexpovstin:
@lead-onym It is necessary to create a short, concise step-by-step plan for the qa tester (up to 3 devices on Android) according to this PR, in order to analyze critical and interface errors.

dev-onym:
Posted the QA plan above and created the marker commit `31ee42c`. Plan is regression-only since this PR ships pure data-type code with no UI surface, and is sized to 3 Android devices covering API-level and form-factor spread.

PR #142, opened 2026-05-19
feat(chats): message persistence layer (PR 2/3 chat stack)

lead-onym:
**Stacked on #141.** Second PR in the Android chat stack. Adds on-disk persistence + a per-group reactive snapshots stream for chat messages. **Mirrors `GroupStore` / `GroupRepository` exactly** so the chat thread UI (later PRs) can subscribe to one group's messages the same way `ChatsScreen` subscribes to the group list. Mirrors [onym-ios PR #148](https://github.com/onymchat/onym-ios/pull/148).

## What's in here

| File | Role |
|---|---|
| `ChatMessage` | Domain type (id, groupId, owner, sender BLS hex, body, sentAt, direction, status, groupType) |
| `MessageDirection` / `MessageStatus` | enums: `INCOMING` / `OUTGOING` ; `PENDING` / `SENT` / `FAILED` / `RECEIVED` |
| `PersistedMessage` | `@Entity`. Plain columns (id / groupId / ownerIdentityId / sentAt / direction / status / groupType) for querying + sort, indexed on `groupId` and `ownerIdentityId`; sender BLS hex and body encrypted via `StorageEncryption` |
| `MessageDao` | Room DAO with the 6 verbs the repo needs (`listForOwnerAndGroup`, `findById`, `insert`, `updateStatus`, `deleteForOwner`, `deleteForGroup`) |
| `MessageDatabase` | **Separate `@Database` + separate `.db` file** from `GroupDatabase` so a schema bump in either domain doesn't force a wipe of the other. The `GroupDatabase` doc-comment predicted this exact split: *"each persistence domain (invitations, groups, messages later) gets its own `@Database`"* |
| `MessageStore` | Async protocol seam |
| `RoomMessageStore` | Wraps encryption at the boundary, tolerant decode (skips corrupt rows), hot-path `updateStatus` skips the encryption round-trip |
| `MessageRepository` | Per-group `MutableStateFlow` cached lazily on first `snapshots(groupId)` call. Chat thread reads one group at a time, so hydrating the whole table at launch would be wasted I/O |
| `InMemoryMessageStore` (sharedTest) | Reusable fake — same pattern as `InMemoryGroupStore` |

## Design notes

- Encryption split matches `PersistedGroup`: queryable / sort columns stay plain, content is encrypted. `(groupId, ownerIdentityId)` get composite indexing via two single-column indices — volumes don't justify a composite index yet.
- `MessageRepository.removeForOwner` refreshes each cached group from the store rather than wiping in-memory — other-identity rows in the same group must survive (tested in `removeForOwner_clearsCachedFlowsForActiveIdentity_butPreservesOtherIdentityRowsInStore`).
- `updateStatus` is a hot path for the outgoing pipeline (`PENDING` → `SENT` / `FAILED`) so we don't round-trip the whole row through the encryption boundary on every relay ACK.
- Identity switch wipes the in-memory cache (via `start()`); the new identity's groups refill on demand. Mirrors the cascade semantics of `GroupRepository`.

## Divergence from iOS

- **`actor MessageRepository` → `Mutex + StateFlow`.** Compose subscribes to `StateFlow` directly via `collectAsStateWithLifecycle()` without per-view ceremony — same reason `GroupRepository` made this swap on day one.
- **`AsyncStream snapshots(groupID:)` → `StateFlow snapshots(groupId)`** backed by a `Map<String, MutableStateFlow<List<ChatMessage>>>` lazy cache. Same lazy-per-group semantics as iOS PR #148.
- **SwiftData `@Model` → Room `@Entity` + DAO**, with a separate `@Database` so the groups and messages schemas migrate independently.

## Out of scope (deferred)

- No wiring to `IncomingMessageDispatcher` or any send path — that lands in PR A4.
- No `OnymApplication` wiring (the `MessageRepository` constructor exists but is unreferenced).
- No production `MessageDatabase` builder in `OnymApplication` — same deferral.

## Test plan

- [x] `RoomMessageStoreTest` — 10 tests (roundtrip, encryption-at-rest, in-place `updateStatus` skips re-encryption, sort, per-(owner, group) filter, cascade `deleteForOwner`, cascade `deleteForGroup`, tolerant decode, empty case, `updateStatus` unknown-id no-op)
- [x] `MessageRepositoryTest` — 10 tests (initial hydration, empty-when-no-active, shared cache across subscribers, `append` emits, per-group isolation, `updateStatus` emits, `updateStatus` unknown-id no-op, cascade `removeForOwner` preserves other-identity rows, `removeForGroup` drops cache, identity switch wipes cache)
- [x] Full `:app:testDebugUnitTest` suite — **461 / 461 pass**, 0 failures, 0 errors, 0 regressions

## Plan context

PR 2 of 3 in the Android chat stack. **PR A1** (#141) added the wire-format payload. **PR A3 next** plumbs the Ed25519 sending pubkey through `MemberProfile` so PR A4's chat dispatcher can verify envelope signatures against the claimed BLS sender.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/chats/ChatMessage.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatMessage.kt
@@ -1,0 +1,46 @@
+package app.onym.android.chats
+
+import app.onym.android.chain.SepGroupType
+import java.util.UUID
+
+/**
+ * Domain shape of one chat message, as the Android app understands
+ * it. Mirrors `ChatMessage.swift` from onym-ios PR #148.
+ *
+ * Two kinds of fields:
+ *  - **Identifying / queryable** ([id], [groupId], [ownerIdentityId],
+ *    [sentAtMillis], [direction], [status], [groupType]) stay plain
+ *    on disk so the DAO can sort and filter without round-tripping
+ *    through the encryption boundary.
+ *  - **Sensitive** ([senderBlsPubkeyHex], [body]) ride encrypted —
+ *    see [PersistedMessage] for the column split.
+ *
+ * [direction] is the local-view distinction (incoming = received
+ * from someone else, outgoing = sent by this device); [status] is
+ * the lifecycle stage (pending → sent / failed, or received). Kept
+ * separate so a `PENDING / OUTGOING` row can survive a relaunch
+ * during a send.
+ */
+data class ChatMessage(
+    /** Stable per-message id. Matches
+     *  [ChatMessagePayload.messageId] on the wire so the receiver
+     *  can dedup re-deliveries. */
+    val id: UUID,
+    /** 64-char lowercase hex of the 32-byte group ID. Matches
+     *  [app.onym.android.group.ChatGroup.id]. */
+    val groupId: String,
+    /** [app.onym.android.identity.IdentityId.value]. Drives the
+     *  per-identity filter on [MessageRepository.snapshots]. */
+    val ownerIdentityId: String,
+    /** 96-char lowercase BLS pubkey hex. Matches
+     *  [app.onym.android.group.ChatGroup.memberProfiles] keying so
+     *  the chat screen can look up the sender's alias with one
+     *  dictionary read. */
+    val senderBlsPubkeyHex: String,
+    val body: String,
+    /** Milliseconds since Unix epoch. */
+    val sentAtMillis: Long,
+    val direction: MessageDirection,
+    val status: MessageStatus,
+    val groupType: SepGroupType,
+)
--- a/app/src/main/kotlin/app/onym/android/chats/MessageDao.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/MessageDao.kt
@@ -1,0 +1,66 @@
+package app.onym.android.chats
+
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+
+/**
+ * Room DAO over [PersistedMessage]. Mirrors the verb set the chat
+ * surface needs: per-(owner, group) list for the chat thread,
+ * insert, status update on the hot path, plus the cascade-delete
+ * variants for identity removal and group deletion.
+ *
+ * Suspend throughout; Room's default single-threaded executor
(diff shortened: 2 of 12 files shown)


dev-onym (commented):
COMMENT — Solid design, mirrors GroupRepository pattern faithfully and tests are thorough. A couple of small correctness/cleanliness items worth a look before PR A4 wires this in.

- `MessageRepository.kt:62-71` — `start()` doc says "Idempotent — safe to call from `onCreate`," but each call launches a new collector on the active-identity flow. Multiple invocations (e.g. across `onCreate` after a config change, or if the caller re-invokes defensively) leak collectors that all wipe `perGroup` on every identity change. Either guard with a `started` flag or document that it must be called exactly once.
- `MessageRepository.kt:99-104` — `if (replaced !== current && replaced != current)` is misleading: `List.map` always returns a new instance, so `replaced !== current` is always true. The effective check is `replaced != current` alone. Drop the reference check or rewrite using `any { it.id == id }` before allocating a new list to actually skip the `map` allocation on unrelated groups (the comment's stated goal).
- `MessageRepository.kt:90` — `append` refreshes via `message.ownerIdentityId` rather than `identity.currentIdentityId.value`. In the steady state these match, but if the active identity has changed between caller-side construction and the `withLock` reaching this line, the cached flow would briefly surface the wrong-identity rows. Using the active identity here (and skipping the refresh if it doesn't match `message.ownerIdentityId`) is safer and matches the cascade-semantics doc.
- `RoomMessageStore.kt:78-87` — `catch (_: Throwable)` in `tryDecryptString` and the `UUID.fromString` decode swallow `OutOfMemoryError` / `StackOverflowError` / `CancellationException`. Narrow to `Exception` (or `GeneralSecurityException` / `IllegalArgumentException`) so VM-level errors and coroutine cancellation propagate.
- `RoomMessageStoreTest.kt:228-230` — `assertNullOk` is unused (the `@Suppress("unused")` is a tell). Either delete or use it in the empty-case test.
- Nit: `MessageStore.deleteForOwner` / `deleteForGroup` return `Int` but `MessageRepository.removeForOwner` / `removeForGroup` discard the count. Either propagate (matches "so the caller can log cleanup size" intent in the DAO doc) or change the store signatures to `Unit`.

PR #143, opened 2026-05-19
feat(identity): plumb Ed25519 sender pubkey through member profiles (PR 3/3 chat stack)

lead-onym:
**Stacked on #142.** Insider-spoofing fix. Plumbs the per-member Ed25519 sending pubkey (== `Identity.stellarPublicKey`, the key that already signs every `SealedEnvelope`) end-to-end through the join + invite flows. **Sets up PR A4's chat dispatcher to verify a chat message's envelope signature against the claimed BLS sender.** Without this, a group member could forge another member's `senderBlsPubkeyHex` claim — the BLS pubkey alone is a claim, not a proof. Mirrors [onym-ios PR #149](https://github.com/onymchat/onym-ios/pull/149).

## Wire compatibility — hard cutover, no migration window

Pre-release, no installed base of join requests or member announcements on the chat path. All three new wire fields are **required** at construction; receivers throw at decode if missing or wrong-sized.

| Type | New field | Wire key |
|---|---|---|
| `MemberProfile` | `sendingPubkey: ByteArray` | `sending_pubkey` |
| `MemberAnnouncementPayload.AnnouncedMember` | `sendingPub: ByteArray` | `sending_pub` |
| `JoinRequestPayload` | `joinerSendingPublicKey: ByteArray` | `joiner_sending_pub` |
| `IdentitySummary` | `sendingPublicKey: ByteArray` (internal projection only) | n/a |

All three wire-shipped types validate `size == 32` at decode via `init { require(...) }` — kotlinx-serialization's synthesized data-class serializer calls the primary constructor, so `init` blocks run during decode (the pattern is already proven by `MemberAnnouncementPayload.AnnouncedMember.init` at lines 103–110 of the pre-PR file).

### Bonus length check on `MemberProfile.inboxPublicKey`

The pre-PR `MemberProfile` had **no length validation at all** — a 31-byte wire `inbox_public_key` would silently decode to an unusable X25519 key. Added `require(inboxPublicKey.size == 32)` alongside the new `sendingPubkey` check. Same one-line fix, same boundary; mirrors iOS PR #149's third commit "fix(MemberProfile): validate field lengths at decode boundary".

## Population sites

- **`JoinRequestSender`** — joiner ships their own `Identity.stellarPublicKey` to the admin on every outbound request.
- **`JoinRequestApprover`**:
  - `PendingRequest` carries the new `joinerSendingPublicKey` through the approval UI.
  - `decode(IntroRequest)` reads it from the wire-decoded payload.
[...]

PR #144, opened 2026-05-19
feat(chats): wire chat-message transport directions (PR 4/5 chat stack)

lead-onym:
**Stacked on #143.** Both directions wired in one PR — symmetric, paired because they share `ChatMessagePayload`. With PRs A1–A3 underneath, two users in a Tyranny group can now exchange encrypted text end-to-end at the data + transport layer. **No UI yet.** Mirrors [onym-ios PR #150](https://github.com/onymchat/onym-ios/pull/150).

## Receive — `IncomingMessageDispatcher` chat-message branch

New third fast path on `dispatch()`: try-decode plaintext as a `ChatMessagePayload`. On match, run the trust chain:

1. **Envelope must have been signed** — guard `envelope.senderEd25519PublicKey != null`. Anonymous chat messages aren't part of the V1 trust model.
2. **Group lookup** — payload's `groupId` must match a local `ChatGroup` owned by the receiving identity. Routes via the new `GroupRepository.findForOwner(ownerIdentityId, groupIdHex)` which **goes through the store directly, bypassing the active-identity filter on `snapshots`**. This is the multi-identity correctness fix: each identity's inbox fan-out supplies the right `ownerIdentityId`, so a message addressed to a non-active identity still lands in the correct group.
3. **Sender lookup** — payload's `sender_bls_pubkey_hex` must be in the group's `memberProfiles`.
4. **Insider-spoof check** — envelope's verified Ed25519 signer must equal that member's stored `MemberProfile.sendingPubkey`. Closes the gap PR A3 (#143) set up; without this, any group member could write another's BLS hex into the payload and the receiver would mis-attribute.
5. **Variant gate** — variant must match the group's governance type.
6. **Persist** via `MessageRepository.append` — idempotent on `ChatMessage.id` (the `MessageDao @Insert` flipped to `OnConflictStrategy.IGNORE` returning `Long`; `-1L` surfaces as `false` at the store seam) so Nostr re-delivery is a no-op.

Dispatcher gains an optional `messageRepository` 4th param. Production wiring in `OnymApplication` instantiates a `MessageDatabase` + `MessageRepository` alongside the existing `GroupDatabase` + `GroupRepository` with the same on-disk-or-in-memory fallback.

## Send — `SendMessageInteractor`

Optimistic-insert pipeline mirroring `CreateGroupInteractor`'s invitation send loop:

1. Validate (identity loaded, group exists, current identity is in roster, body non-empty, variant matches governance).
2. Mint `messageId` + payload, persist locally as `PENDING` so the chat thread shows the bubble immediately.
3. Fan out one sealed envelope per **other** member's inbox key — best-effort, swallow per-recipient failures.
4. Flip status: `SENT` if any relay accepted (or no recipients — note-to-self is local-only), `FAILED` if all sends rejected.

Today only the `Tyranny` variant ships; other group types throw `UnsupportedGroupType` until their variants come online.

**Takes `ActiveIdentityProvider` + `identitiesFlow` + `InvitationEnvelopeSealer` + `InboxTransport`** instead of the concrete `IdentityRepository` so the interactor is unit-testable without standing up the real keychain. Production wiring passes `identityRepository` for all three of the identity slots (it implements both `ActiveIdentityProvider` and `InvitationEnvelopeSealer` and exposes `identities` as a `StateFlow<List<IdentitySummary>>`).

## Tests

- **`IncomingMessageDispatcherChatMessageTest`** (9): happy path, unknown group / sender / signature mismatch / missing signature / wrong owner identity / **non-active identity (proves multi-identity)** / duplicate messageID idempotency / variant-vs-governance mismatch.
- **`SendMessageInteractorTest`** (9): persist + fan out marks SENT, skip self recipient, zero peers (note-to-self), all-fail marks FAILED, send-throws marks FAILED, partial success marks SENT, `unknownGroup` / `senderNotAMember` / `emptyBody` validation paths, `unsupportedGovernance` throws.
- Full `:app:testDebugUnitTest` suite — **490 / 490 pass**, 0 failures, 0 errors, 0 regressions.

## Multi-identity correctness (called out)

A specific test (`chatMessage_persistsForNonActiveIdentity`) seeds two identities, leaves Alice active, and dispatches a message addressed to Bob's identity. Asserts the message lands in Bob's group regardless. The dispatcher routes by the `ownerIdentityId` parameter through `GroupRepository.findForOwner` (which bypasses the active-identity snapshot filter), so inbox fan-out per identity does the right thing.

## Divergence from iOS

- **`actor SendMessageInteractor` → plain `class`** + `withContext(ioDispatcher)`. Same serialization guarantees via the dispatcher hop.
- **`identity: IdentityRepository` → split interfaces** (`ActiveIdentityProvider` + `StateFlow<List<IdentitySummary>>` + `InvitationEnvelopeSealer`). Android-side testability — no `actor`-style protocol on the concrete repo to stub against.
- **`messageRepository: MessageRepository` non-optional on iOS → optional on Android** (`?= null`). Lets the pre-existing `IncomingMessageDispatcherTest` (and the announcement / invitation fast paths) keep working without threading a `MessageRepository` through every construction site.

## Out of scope

UI — that's PR A5 (Compose chat-thread screen).

## Plan context

PR 4 of 5 in the Android chat stack. **PR A1** (#141) payload • **PR A2** (#142) persistence • **PR A3** (#143) Ed25519 sender pubkey. **PR A5 next**: Compose chat-thread screen + nav rewire (members list moves one tap deeper behind an info button).

### DIFF
--- a/app/src/main/kotlin/app/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/app/onym/android/OnymApplication.kt
@@ -310,6 +310,36 @@ class OnymApplication : Application() {
         groupRepository.start()
         applicationScope.launch { groupRepository.reload() }
 
+        // Message persistence (PR A2 + A4). Separate @Database from
+        // GroupDatabase so a schema bump in either domain doesn't
+        // force a wipe of the other — exact mirror of the
+        // GroupDatabase comment's prediction. Same in-memory fallback
+        // posture as the groups DB: a corrupt on-disk file degrades
+        // to ephemeral storage rather than refusing to launch.
+        val messageDatabase = try {
+            androidx.room.Room.databaseBuilder(
+                applicationContext,
+                app.onym.android.chats.MessageDatabase::class.java,
+                "app.onym.android.messages",
+            )
+                .fallbackToDestructiveMigration()
+                .build()
+        } catch (_: Throwable) {
+            androidx.room.Room.inMemoryDatabaseBuilder(
+                applicationContext,
+                app.onym.android.chats.MessageDatabase::class.java,
+            ).build()
+        }
+        val messageRepository = app.onym.android.chats.MessageRepository(
+            store = app.onym.android.chats.RoomMessageStore(
+                dao = messageDatabase.messageDao(),
+                encryption = storageEncryption,
+            ),
+            identity = identityRepository,
+            scope = applicationScope,
+        )
+        messageRepository.start()
+
         // App-wide network preference (PR-C follow-up). Constructed
         // here (early) so PR-89's chainStateReader can read it.
         val networkPreference = DataStoreNetworkPreferenceProvider(
@@ -385,6 +415,7 @@ class OnymApplication : Application() {
             envelopeDecrypter = identityRepository,
             groupRepository = groupRepository,
             invitationsRepository = invitationsRepository,
+            messageRepository = messageRepository,
             identitiesFlow = identityRepository.identities,
             chainState = chainStateReader,
         )
--- a/app/src/main/kotlin/app/onym/android/chats/MessageDao.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/MessageDao.kt
@@ -39,12 +39,14 @@ interface MessageDao {
     @Query("SELECT * FROM messages WHERE id = :id LIMIT 1")
     suspend fun findById(id: String): PersistedMessage?
 
-    /** Pure insert. Throws on PK conflict — the store's caller has
-     *  already routed through dedup (the [PersistedMessage.id] is
-     *  the wire-stable message UUID, so re-delivery must be filtered
-     *  upstream of this DAO). */
-    @Insert(onConflict = OnConflictStrategy.ABORT)
-    suspend fun insert(record: PersistedMessage)
+    /** Idempotent on [PersistedMessage.id]: a re-delivery of the
+     *  same wire message is a silent no-op rather than a thrown
+     *  exception, mirroring the dispatcher's "drop duplicates"
+     *  contract. Returns the row id on insert and `-1` when the
+     *  conflict strategy ignored the write — callers expose that as
+     *  `Boolean` at the [MessageStore] seam. */
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
+    suspend fun insert(record: PersistedMessage): Long
 
     /** Hot path: flip `PENDING` → `SENT` / `FAILED` (or any other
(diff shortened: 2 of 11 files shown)


dev-onym (commented):
COMMENT — solid, mirrors iOS faithfully; flagging a few small issues, none blocking.

- `IncomingMessageDispatcher.kt:383` — `@Suppress("UNUSED_VARIABLE") val _receivedAt = receivedAt` is dead code; the comment promises future use but YAGNI says drop the assignment (parameter still legitimately documents the dispatcher's contract).
- `IncomingMessageDispatcher.kt:359-360` — when `messageRepository` is null, a successfully-decoded chat payload is silently dropped (no fall-through to invitations queue, no log). Production wires it via `OnymApplication.kt:418`, but any future construction site that forgets it will lose messages without trace. Worth at least a `kotlin.error`/log gate, or removing the `?= null` default now that production wires it.
- `SendMessageInteractor.kt:68` — `body.isEmpty()` accepts whitespace-only bodies ("   "). iOS does the same so this preserves parity, but a `isBlank()` check would be friendlier; if intentional, ignore.
- `SendMessageInteractorTest.kt:122-142` — test name `send_partialSuccess_marksSentWhenAtLeastOneAccepted` is misleading: both sends succeed (same `acceptedBy` applies to all recipients), so the *partial*-success branch (some accept, some reject) isn't actually covered. The PR body claims this case is tested — either rename to `multipleRecipients_allAccepted` or extend `ConfigurableInboxTransport` to script per-recipient receipts.
- `SendMessageInteractor.kt:114` — optimistic `messageRepository.append(pending)` has no try/catch; an exception there propagates after validation but BEFORE any wire send, leaving no row at all. Fine (caller sees the exception), just noting it's the only error path that doesn't flip to `FAILED`.
- Trust chain at `IncomingMessageDispatcher.kt:353-385` is correctly ordered: signer-required → owner-scoped group lookup → member lookup → Ed25519/sendingPubkey equality → variant gate → idempotent append. The multi-identity routing via `findForOwner` (bypassing `snapshots`' active-identity filter) is the right fix and is well-covered by `chatMessage_persistsForNonActiveIdentity`.
- `MessageDao.kt:42-49` `OnConflictStrategy.IGNORE` returning `Long` and surfacing as `Boolean` at `RoomMessageStore.kt:34-38` / `InMemoryMessageStore.kt:37-42` is consistent; `MessageRepository.append` skipping the cache refresh on duplicate is a nice extra (avoids storm).

PR #145, opened 2026-05-19
feat(chats): Compose chat-thread shell + nav rewire (PR 5/5 chat stack)

lead-onym:
**Stacked on #144.** First user-visible PR of the Android chat stack. Tapping a group in the Chats tab now opens a Compose `ChatThreadScreen` instead of the members roster. The members list moves one tap deeper, behind the thread's info button. Mirrors [onym-ios PR #151](https://github.com/onymchat/onym-ios/pull/151).

## What's in here

**`ChatThreadScreen.kt`** (Compose, Material 3)
- `Scaffold` + `TopAppBar` mirroring `ChatMembersScreen`'s shape: back arrow on the left, group name (or `"Chat"` fallback) in the center, info action on the right.
- Empty body placeholder where the message list will render in the next PR.
- Input panel placeholder where the input field + send button will land in the next PR.
- Test tags on every interactive surface (`chat_thread.back_button`, `chat_thread.info_button`, `chat_thread.title`, `chat_thread.body`, `chat_thread.placeholder_text`, `chat_thread.input_panel_placeholder`, `chat_thread.missing`) so the future UI iteration has stable hooks.
- Missing-group state when the path-arg id doesn't resolve.

**`ChatThreadViewModel.kt`**
- `group: StateFlow<ChatGroup?>` resolves by id from `GroupRepository.snapshots`; `null` when missing.
- `messages: StateFlow<List<ChatMessage>>` drains `MessageRepository.snapshots(groupId)` into a hot flow the screen collects.
- `send(body)`: trims + dispatches through the captured `SendMessageInteractor`; routes errors into `lastSendError`; flips `sendInFlight` around the call. Empty / whitespace bodies short-circuit before reaching the interactor.
- `clearError()` / `sendInFlight` / `lastSendError` surface state the next PR's input panel will bind to without any VM changes.

**Wiring**
- `AppDependencies` gains `makeChatThreadViewModel: (groupId) -> ChatThreadViewModel`.
- `OnymApplication.buildDependencies()` constructs a fresh `SendMessageInteractor` per VM (interactor is stateless; the closure captures `identityRepository` / `messageRepository` / `inboxTransport` so the VM constructor stays narrow).
- `RootScreen.NavHost`: new `composable("chat_thread/{groupId}")` destination; `ChatsScreen.onOpenChat` rewired from `"chat_members/$groupId"` to `"chat_thread/$groupId"`. The info button on the thread navigates to the existing `chat_members` destination (one tap deeper, same screen as before).

## Divergence from iOS

iOS PR #151 wraps a UIKit `ChatThreadViewController` in a SwiftUI `UIViewControllerRepresentable` bridge because SwiftUI's nav primitives don't cover the diffable-data-source pattern UIKit needed for the message list. **Compose's `LazyColumn` + the existing M3 `TopAppBar` primitives cover what iOS reached for UIKit for, so the Android twin is a single Compose composable with no bridge.**

## Out of scope

- Message bubble rendering (next PR).
- Input panel + send button + keyboard handling (next PR).
- Auto-scroll to bottom on new message arrival (next PR).
- Routing the post-join-success callback from `JoinScreen` to the chat thread of the freshly-materialized group (separate polish PR — the existing `RootScreen` comment at the `join_invite` destination already calls out this future hook).

## Tests

- **`ChatThreadViewModelTest`** (9): group resolves by id, group `null` when missing, messages mirror `MessageRepository` snapshots, messages emit on `append`, send invokes interactor with trimmed body, send empty/whitespace short-circuited, send interactor throws routes into `lastSendError`, `clearError` resets, send success clears previous error.
- Full `:app:testDebugUnitTest` suite — **499 / 499 pass**, 0 failures, 0 errors, 0 regressions.
- `assembleDebug` builds cleanly (Compose code paths verified beyond unit-test compile).

Layout, colors, and look-and-feel are not under test. Launch the app, tap a group: top bar shows the group name, back arrow pops, info button pushes the members screen, table view fills the body, input placeholder sits at the bottom safe area.

## Plan context

PR 5 of 5 in the Android chat stack. **PR A1** (#141) payload • **PR A2** (#142) persistence • **PR A3** (#143) Ed25519 sender pubkey • **PR A4** (#144) dispatcher + send interactor. Next slice (separate PR): message bubble rendering — Lazy column + custom bubble composable, subscribe to `messages` already exposed here, auto-scroll on append.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/AppDependencies.kt
+++ b/app/src/main/kotlin/app/onym/android/AppDependencies.kt
@@ -2,6 +2,7 @@ package app.onym.android
 
 import androidx.fragment.app.FragmentActivity
 import app.onym.android.chain.NetworkPreferenceProvider
+import app.onym.android.chats.ChatThreadViewModel
 import app.onym.android.chats.ChatsViewModel
 import app.onym.android.group.CreateGroupViewModel
 import app.onym.android.recovery.RecoveryPhraseBackupViewModel
@@ -45,6 +46,12 @@ class AppDependencies(
     /** Chats tab — read-only view over [app.onym.android.group.GroupRepository.snapshots].
      *  Mirrors `makeChatsFlow` from onym-ios PR #30. */
     val makeChatsViewModel: () -> ChatsViewModel,
+    /** Chat-thread screen factory — takes the path-arg `groupId` so
+     *  the VM can subscribe to that group's [app.onym.android.chats.MessageRepository.snapshots]
+     *  stream and dispatch sends via the captured [SendMessageInteractor].
+     *  Mirrors the UIViewControllerRepresentable bridge factory from
+     *  onym-ios PR #151, Android-idiomatic types. */
+    val makeChatThreadViewModel: (groupId: String) -> ChatThreadViewModel,
     /** Settings → Identities — multi-identity management (PR-5). */
     val makeIdentitiesViewModel: () -> app.onym.android.identity.IdentitiesViewModel,
     /** Post-create deeplink invite share (deeplink-invite PR-5). */
--- a/app/src/main/kotlin/app/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/app/onym/android/OnymApplication.kt
@@ -567,6 +567,27 @@ class OnymApplication : Application() {
             makeChatsViewModel = {
                 app.onym.android.chats.ChatsViewModel(repository = groupRepository)
             },
+            makeChatThreadViewModel = { groupId ->
+                // PR A5: per-thread VM. The SendMessageInteractor is
+                // built fresh per VM — it's stateless and capturing
+                // the closure here keeps the interactor's
+                // dependencies (identity, transport, message repo)
+                // out of the VM's own constructor.
+                val sender = app.onym.android.chats.SendMessageInteractor(
+                    activeIdentity = identityRepository,
+                    identitiesFlow = identityRepository.identities,
+                    envelopeSealer = identityRepository,
+                    groupRepository = groupRepository,
+                    messageRepository = messageRepository,
+                    inboxTransport = inboxTransport,
+                )
+                app.onym.android.chats.ChatThreadViewModel(
+                    groupId = groupId,
+                    groupRepository = groupRepository,
+                    messageRepository = messageRepository,
+                    sendMessage = sender::send,
+                )
+            },
             makeIdentitiesViewModel = {
                 app.onym.android.identity.IdentitiesViewModel(identity = identityRepository)
             },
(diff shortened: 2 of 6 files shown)


dev-onym (commented):
**COMMENT** — well-structured slice with thorough tests; one structured-concurrency bug worth fixing before merge.

- `ChatThreadViewModel.kt:86` — `catch (e: Throwable)` swallows `CancellationException`. When the VM is cleared mid-send, `viewModelScope` cancels the launched coroutine; the suspend call throws `CancellationException`, which is caught here and recorded as a "send error" instead of propagating. Either rethrow CE explicitly (`if (e is CancellationException) throw e`) or catch only `Exception`. Same Throwable catch also means OOM/AssertionError become user-facing toast text.
- `ChatThreadViewModel.kt:53-59` — `messages` uses an eager `MutableStateFlow + .also { launch { collect } }` pattern while `group` (line 41-48) uses `stateIn(WhileSubscribed)`. Inconsistent lifecycle (messages collects forever until VM clear; group times out at 5s after last subscriber). If intentional, fine — but the more idiomatic `messageRepository.snapshots(groupId).stateIn(viewModelScope, WhileSubscribed(5_000), emptyList())` matches the `group` pattern.
- `ChatThreadViewModel.kt:46-47` — eager `initialValue = groupRepository.snapshots.value.firstOrNull{...}` duplicates the `map` logic and reads `.value` at VM-construction time on the calling thread. `stateIn` will populate from the upstream anyway; `initialValue = null` is simpler and matches the doc'd nullable contract.
- `ChatThreadViewModel.kt:11` — `flatMapLatest` imported but unused.
- `ChatThreadViewModelTest.kt:64` — typo `messages_minorOfMessageRepositorySnapshots` (s/minor/mirror/).
- `ChatThreadViewModelTest.kt:166` — `SendMessageError.EmptyBody` thrown from the fake to exercise `clearError`, but VM short-circuits empty bodies before the interactor — the error name is misleading here (test still passes since body is `"hi"`). Pick a different variant for clarity.
- `OnymApplication.kt:570-587` — constructing a new `SendMessageInteractor` per VM is fine if truly stateless, but the closure captures four repositories per nav-entry; a single shared interactor instance at app construction (used by the factory) would be cheaper and equally narrow on the VM constructor. Minor.
- `ChatThreadScreen.kt:128,154` — placeholder strings ("rendering lands in the next PR", "Message input lands in the next PR") are dev-facing copy in a user-visible build. Confirmed scoped to next PR per the description; flagging so it doesn't ship if the next PR slips.
- Nav: `RootScreen.kt:177` silently `return@composable` on missing `groupId` arg — acceptable since the route requires the path arg, but a log line would help diagnose any future deep-link bug.

PR #146, opened 2026-05-19
feat(chats): message list rendering (PR 6/N chat stack)

lead-onym:
**Stacked on #145.** Read-only chat scrollback. Messages from `MessageRepository` flow through the VM into a `LazyColumn` and render as direction-styled bubbles. **No input or send yet** — that's the next PR. Mirrors [onym-ios PR #152](https://github.com/onymchat/onym-ios/pull/152).

### `ChatBubble.kt` (new Compose composable)

Two visual modes driven by `ChatMessage.direction`:
- `OUTGOING` → `colorScheme.primary` fill, `onPrimary` text, trailing-aligned.
- `INCOMING` → `colorScheme.surfaceVariant` fill, `onSurfaceVariant` text, leading-aligned.

Bubble max width is capped at 75% of the parent row via `BoxWithConstraints` — tracks the real measurement, no hardcoded `Dp` ceiling.

### `ChatThreadScreen.kt`

Body placeholder replaced with a real `LazyColumn(items, key = { it.id })`. Compose's stable-key `items()` covers the same data-diff concerns iOS reached for `UITableViewDiffableDataSource` for; status flips (`PENDING → SENT / FAILED`) re-render the bubble naturally because `ChatMessage.equals` compares all fields — no explicit diffable-identity widening needed.

**Auto-scroll** mirrors iOS:
- **Cold open** (first render with any messages): non-animated jump to the latest message via `scrollToItem(lastIndex)`. Guarded by a remembered `hasInitialScrolled` flag.
- **Subsequent appends**: only `animateScrollToItem(lastIndex)` if the user was already near the bottom (threshold = 2 items from the end). Reading older messages doesn't get hijacked.

The near-bottom check rides on a `derivedStateOf` over `listState.layoutInfo` so it only recomputes when the visible window or item count changes.

**Defensive sort** at the body composable level — `sortedBy { it.sentAtMillis }`. The repository's contract is ascending, but a defensive sort protects against a future caller / test stub regressing the order. Mirrors the iOS PR #152 fixup.

Empty-thread state ("No messages yet. Say hi.") replaces the prior PR's "Bubbles render in the next PR" placeholder. Input panel stays a placeholder — that's the next PR.

## Divergence from iOS

- **UIKit `ChatBubbleCell` + diffable data source → Compose `LazyColumn(items, key = { it.id })`.** No cell reuse, no Auto-Layout constraint pairs, no diffable-identity reconfigure for status flips — Compose's recomposition over the data class handles all three for free. The iOS fixup commit's biggest item (resolving the Auto-Layout conflict between always-active gap constraints and toggled-pin alignment) doesn't exist as a problem on Compose.
- **iOS `preferredFont(forTextStyle: .body)` → `MaterialTheme.typography.bodyMedium`.** Both pick up the system text-size accessibility setting; same a11y posture by default.
- **Pixel-offset `contentOffset` near-bottom check → item-index near-bottom check.** Compose's `LazyListState` doesn't expose pixel offsets at the public API, so the heuristic is expressed in item indices. Threshold = 2 items ≈ the iOS 100pt threshold for a typical bubble row height.

## What's NOT in here

- Input UI / send wiring (next PR).
- Keyboard avoidance (next PR).
- "↓ new messages" pill — deferred polish; for now, messages that arrive while the user is scrolled up appear without disrupting their position (matches iOS's same deferral).
- Avatars, timestamps, status glyphs, emoji rendering (later PRs).

## Tests

- **`ChatThreadAutoScrollTest`** (7): pure-policy tests for `isNearBottom` — empty list is true, last index visible is true, scrolled up is false, threshold edges (one-from-end is near; two-from-end is not), no-visible-index-yet is conservative-false, custom-threshold knob.
- The visual rendering and auto-scroll integration are NOT under test (matches iOS's "Visual correctness — verify in simulator" posture); the unit-testable policy lives in the pure `isNearBottom` function which the composable wires via `derivedStateOf`.
- Full `:app:testDebugUnitTest` suite — **506 / 506 pass**, 0 failures, 0 errors, 0 regressions.
- `assembleDebug` builds cleanly.

Bubble shape, colors, scroll behavior: launch the app, open a chat. Easiest end-to-end demo: two devices in the same Tyranny group, send a message from one, watch it land on the other. The receive path is already covered by `IncomingMessageDispatcherChatMessageTest` in #144.

## Plan context

PR 6 of N in the Android chat stack. **PR A1** (#141) payload • **PR A2** (#142) persistence • **PR A3** (#143) Ed25519 sender pubkey • **PR A4** (#144) dispatcher + send interactor • **PR A5** (#145) Compose chat-thread shell. **Next PR**: input panel + send wiring — `TextField` + send button + keyboard handling, hook the Send button to the `ChatThreadViewModel.send` method which already routes through `SendMessageInteractor`.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
@@ -1,0 +1,95 @@
+package app.onym.android.chats
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxWithConstraints
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.unit.dp
+
+/**
+ * Chat-thread bubble. Two visual modes driven by [ChatMessage.direction]:
+ *
+ *  - [MessageDirection.OUTGOING] — `colorScheme.primary` fill,
+ *    `colorScheme.onPrimary` text, trailing-aligned.
+ *  - [MessageDirection.INCOMING] — `colorScheme.surfaceVariant`
+ *    fill, `colorScheme.onSurfaceVariant` text, leading-aligned.
+ *
+ * Bubble max width is capped at [maxWidthFraction] of the parent
+ * row (default 75%) so a long monologue doesn't reach the opposite
+ * edge. The cap is resolved via `BoxWithConstraints` so it tracks
+ * the real parent measurement — no hardcoded `Dp` ceiling.
+ *
+ * Compose's natural recomposition over the [ChatMessage] data class
+ * means a status flip (PENDING → SENT / FAILED) re-renders the
+ * bubble without any explicit diffable-identity widening — the
+ * outer `LazyColumn(items, key = { it.id })` keeps the slot stable;
+ * this body reads the latest fields.
+ *
+ * Status glyphs, timestamps, avatars, emoji rendering are out of
+ * scope for the read-only scrollback PR — they land in later
+ * polish slices.
+ *
+ * Mirrors `ChatBubbleCell.swift` from onym-ios PR #152, Compose-
+ * idiomatic: no constraint pairs, no cell reuse, no diffable data
+ * source — `BoxWithConstraints` + `LazyColumn(items, key = ...)`
+ * cover all three concerns.
+ */
+@Composable
+fun ChatBubble(
+    message: ChatMessage,
+    modifier: Modifier = Modifier,
+    maxWidthFraction: Float = 0.75f,
+) {
+    val isOutgoing = message.direction == MessageDirection.OUTGOING
+    val fill: Color = if (isOutgoing) {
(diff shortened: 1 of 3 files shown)


dev-onym (commented):
**APPROVE**

- `ChatBubble.kt:74` — `BoxWithConstraints` + `widthIn(max = maxWidth * maxWidthFraction)` correctly caps width at 75% of measured parent; Dp×Float arithmetic is valid.
- `ChatThreadScreen.kt:135` — `LaunchedEffect(sortedMessages.size)` keying on size is intentional and consistent with the PR's claim that status flips re-render via `ChatMessage.equals` without disturbing scroll. Note: a same-size list replacement (e.g., full refresh with N→N) won't trigger any scroll-to-bottom, but for the streaming-append model this is fine.
- `ChatThreadScreen.kt:131` — `nearBottom` reads `layoutInfo` *after* the new item is composed; relies on Compose's layout-before-effects ordering. In practice the heuristic still works because when the user was at the prior last item, that item becomes second-to-last and `lastVisibleIndex >= totalItems - 2` stays true on appends of 1. Good.
- `ChatThreadScreen.kt:120` — `remember(messages) { messages.sortedBy { it.sentAtMillis } }` re-sorts on every StateFlow emission (new List identity). O(n log n) on already-sorted input is acceptable; `sortedBy` is stable so ties preserve insertion order.
- `ChatThreadScreen.kt:144` — `scrollToItem(lastIndex)` is also fired on deletions (size change), not just appends. Deletions aren't a current path so this is harmless, but worth noting if delete lands later.
- `ChatThreadScreen.kt:226` — `NEAR_BOTTOM_INDEX_THRESHOLD = 2` matches the test expectations (`lastVisibleIndex >= totalItems - 2` ⇒ last two indices count). Comment in the test ("third-to-last") and assertion (`47 < 48`) are consistent.
- `ChatThreadAutoScrollTest.kt` — pure-policy tests cover empty / at-end / scrolled-up / both threshold edges / no-measure-yet / custom-threshold knob. Good coverage of the unit-testable surface; visual + integration deferred to manual is reasonable.
- `ChatBubble.kt:91` — `BUBBLE_RADIUS` private file constant placement is fine; would be marginally cleaner near the top but not worth changing.

PR #147, opened 2026-05-19
feat(chats): input panel + keyboard avoidance (PR 7/N chat stack)

lead-onym:
**Stacked on #146.** Composer is real now. An `OutlinedTextField` that grows to a 3-line cap, a trailing Send button, and `Modifier.imePadding()` on the body column so the panel rides with the soft keyboard. **No real send wiring yet** — tap clears the field; the body is discarded. Next PR swaps the no-op closure for `viewModel.send` (already wired through `SendMessageInteractor` in PR A4). Mirrors [onym-ios PR #153](https://github.com/onymchat/onym-ios/pull/153).

## `ChatInputPanel`

- `OutlinedTextField` + `FilledIconButton`. Material 3 throughout.
- `"Message"` placeholder via `TextField`'s built-in `placeholder` slot — no manual overlay label needed.
- **Auto-grow** up to `maxLines = 3`, then scrolls internally. Compose's `TextField` does this natively; no intrinsic-size measurement, no `layoutIfNeeded()` flushes (no TextKit-2 staleness bug to fight).
- `onSend(body)` fires with the body trimmed of leading + trailing whitespace. Whitespace-only input never reaches the callback:
  - the `FilledIconButton` is `enabled = canSend` where `canSend` consults the trim, and
  - the click handler re-runs the trim as belt-plus-braces and returns early if the body is `null`.
- Text state is internal via `remember { mutableStateOf("") }`. The composable clears itself after firing `onSend`. A future "draft persistence" PR can hoist the state into the VM.

## `ChatThreadScreen` wire-up

- Replaced `InputPanelPlaceholder()` with `ChatInputPanel(onSend = { _ -> })`.
- Added `Modifier.imePadding()` to the body column. With no keyboard up the padding is zero; when the keyboard rises the panel slides automatically. Same affordance as iOS's `keyboardLayoutGuide`.
- `enableEdgeToEdge()` was already on in `MainActivity` (existing setup); `imePadding()` relies on that.

## Divergence from iOS

- **`UITextView` + manual `intrinsicContentSize` measurement → `OutlinedTextField(maxLines = 3)`.** Compose's `TextField` grows then scrolls natively.
- **`view.keyboardLayoutGuide.topAnchor` → `Modifier.imePadding()`.** Declarative; no notification observers, no frame math.
- **Manual placeholder `UILabel` overlay → `TextField`'s `placeholder` slot.**
- **TextKit-2 `layoutIfNeeded()` workaround doesn't apply** — Compose has no equivalent stale-measurement issue.

## Tests

- **`ChatInputPanelTest`** (7): pure-policy tests for `trimmedSendBody`. Empty / whitespace-only returns `null`; non-empty after trim returns the trimmed body; internal whitespace preserved; unicode round-trips. Plus a small wrapper-test that asserts the Send-enabled derivation matches the `trimmedSendBody-is-non-null` shape the composable uses.
- The visual rendering, keyboard slide animation, and 3-line cap are NOT under test (matches iOS's "Visual correctness — verify in simulator" posture). The unit-testable policy lives in the pure `trimmedSendBody` function.
- Full `:app:testDebugUnitTest` suite — **513 / 513 pass**, 0 failures, 0 errors, 0 regressions.
- `assembleDebug` builds cleanly.

Type-while-keyboard-shown layout, keyboard slide animation, the 3-line growth cap before scrolling, Dynamic Type scaling: launch the app, tap a group, tap the field, type, hit Send → the field should clear and the keyboard stays up.

## Plan context

PR 7 of N in the Android chat stack. **PR A1** (#141) payload • **PR A2** (#142) persistence • **PR A3** (#143) Ed25519 sender pubkey • **PR A4** (#144) dispatcher + send interactor • **PR A5** (#145) Compose chat-thread shell • **PR A6** (#146) message list rendering. **Next PR**: swap `onSend = { _ -> }` for `viewModel.send(body)` — single-line change; the VM already plumbs the call through `SendMessageInteractor` end-to-end.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/chats/ChatInputPanel.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatInputPanel.kt
@@ -1,0 +1,137 @@
+package app.onym.android.chats
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.automirrored.filled.Send
+import androidx.compose.material3.FilledIconButton
+import androidx.compose.material3.Icon
+import androidx.compose.material3.IconButtonDefaults
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.OutlinedTextFieldDefaults
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.input.ImeAction
+import androidx.compose.ui.text.input.KeyboardCapitalization
+import androidx.compose.ui.unit.dp
+
+/**
+ * Chat-thread composer. A `TextField` that grows to a 3-line cap
+ * (then scrolls internally) with a trailing Send button. Tapping
+ * Send fires [onSend] with the body trimmed of leading and trailing
+ * whitespace, then clears the field. Whitespace-only input never
+ * fires the callback — the button stays disabled and the trim guard
+ * in [trimmedSendBody] is the belt-plus-braces check.
+ *
+ * Keyboard avoidance is the parent's responsibility: place this
+ * inside a `Modifier.imePadding()`-bounded layout. With no keyboard
+ * up the padding is zero; when the keyboard rises the parent
+ * automatically slides up — same affordance as iOS's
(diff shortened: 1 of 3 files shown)


dev-onym (commented):
APPROVE
- `app/src/main/kotlin/app/onym/android/chats/ChatInputPanel.kt:3` — unused `import androidx.compose.foundation.background`. Minor lint-only nit; safe to drop.
- `ChatInputPanel.kt:80-97` — `clip(RoundedCornerShape(...))` on the text field's `modifier` is redundant with the `shape = RoundedCornerShape(BUBBLE_CORNER)` arg right below it; `OutlinedTextField` already clips to its shape. Not a bug, just dead pixels.
- `ChatInputPanel.kt:87-90` — `ImeAction.Default` + no `keyboardActions` means Enter inserts a newline rather than firing send. Matches the iOS twin and the 3-line cap, so correct, but worth a confirm on UX intent.
- `ChatInputPanel.kt:62,84` — `enabled` is forwarded to the field but the panel is always rendered with default `enabled = true` from the screen. Fine for now; no action.
- `ChatInputPanel.kt:99-103` — belt-plus-braces `trimmedSendBody` re-check inside `onClick` is sound (guards against a stale `canSend` snapshot under fast taps); good defensive choice.
- `ChatThreadScreen.kt:162` — `imePadding()` after `padding(padding)` is the correct order (scaffold insets first, then IME).
- `ChatThreadScreen.kt:199` — `onSend = { _ -> }` discards the body as documented; next PR's promised one-line swap is plausible.
- Tests: `ChatInputPanelTest.kt` exercises the only unit-testable surface (`trimmedSendBody`); visual/keyboard behavior correctly left to manual verification per the PR body.

PR #148, opened 2026-05-19
feat(chats): wire Send button to viewModel.send (PR 8/N chat stack)

lead-onym:
**Stacked on #147.** End-to-end. Type a message, tap Send, see it appear in the thread immediately and propagate to the other group members. Mirrors [onym-ios PR #154](https://github.com/onymchat/onym-ios/pull/154).

## What's in here

Three-line edit. PR A4 already built `SendMessageInteractor` with the full optimistic-insert / fan-out / status-flip lifecycle, and PR A5 wired it into `ChatThreadViewModel.send`. The body itself is internal to the input panel (PR A7's `remember { mutableStateOf("") }`), so this PR is just the lambda swap: `ChatInputPanel(onSend = onSend)` where `onSend` in the parent is `viewModel::send`. The heavy plumbing was done two PRs ago.

## UX flow

1. User types "hi" → Send button enables (PR A7 trim gate).
2. Tap Send → `ChatInputPanel` forwards the trimmed body to `onSend`.
3. `viewModel.send(body)` → `SendMessageInteractor` inserts a `PENDING` row into `MessageRepository` **before** the network await.
4. `MessageRepository.snapshots` emits → PR A6's `LazyColumn(key = id)` reconciles → bubble appears + auto-scroll triggers.
5. Input clears immediately (panel's own clear-after-onSend logic — snappier than waiting for the await).
6. Network round-trip resolves → interactor flips status to `SENT` / `FAILED`. The status visual indicator on the bubble lands in a later PR.

`SendMessageError` cases (no identity loaded, unknown group, not a member, empty body) all represent invariants that should hold mid-chat-screen and route into `viewModel.lastSendError`, which the UI doesn't surface yet. Network / fan-out failures land as `MessageStatus.FAILED` on the bubble — mirrors iOS PR #154's same swallow-and-defer-to-status posture.

## Divergence from iOS

iOS PR #154 threads `sendMessageInteractor` through `AppDependencies` → `RootView` → `ChatsView` → `ChatThreadView` and exposes `onSendTapped` on the `ChatThreadViewController`. Android's equivalent is already in place from PR A5: the `makeChatThreadViewModel` factory in `OnymApplication` captures the interactor in a closure; the VM exposes `send(body)`. **PR A8 is just the lambda swap in the screen.**

## Tests

The wire-up itself is a one-line lambda in a Compose composable; testing it directly requires Compose UI tests (androidTest). All three layers it composes are already tested by earlier PRs in this stack:

[...]

PR #149, opened 2026-05-19
feat(chats): message status indicator + retry on failed (PR 9/N chat stack)

lead-onym:
**Stacked on #148.** Status glyphs on outgoing bubbles + tap-to-retry on failed. Tapping a failed bubble retries the send with the original message id so receivers dedup against any prior delivery. Mirrors [onym-ios PR #155](https://github.com/onymchat/onym-ios/pull/155).

## `ChatBubble`

Status indicator below the bubble's trailing edge (outgoing only):

| Status | Glyph | Tint |
|---|---|---|
| `PENDING` | `Icons.Outlined.Schedule` | `onSurfaceVariant` (muted) |
| `SENT` | `Icons.Filled.Check` | `onSurfaceVariant` (muted) |
| `FAILED` | `Icons.Filled.ErrorOutline` | `error` (red) |

Incoming hides the indicator (the message arriving is proof of delivery).

**Retry tap**: `Modifier.clickable` on the bubble box only when `canRetry(message) == true` (i.e. `OUTGOING + FAILED`). Compose's conditional-modifier composition replaces iOS's add/remove-`UITapGestureRecognizer` dance on cell reuse.

`statusVisualFor(...)` + `canRetry(...)` are pure functions — unit tests pin the glyph mapping + retry gate without standing up Compose.

## `SendMessageInteractor`

- **Extracted private `fanOut(payload, recipients)`** helper. Same encode + per-recipient seal + transport flow used by both send and the new retry path; one source of truth for the best-effort per-recipient swallow.
- **New `retry(groupId, messageId)`** — silent no-op for unknown / non-failed / non-outgoing / non-Tyranny rows so a stale tap after a status flip can't double-deliver. On a valid retry:
  1. Flip status to `PENDING` immediately (glyph swaps to the in-flight clock).
  2. Re-fan-out with the **original** payload fields (`messageId`, `body`, `sentAtMillis`) so receivers dedup against any prior delivery via the dispatcher's existing `MessageRepository.append` idempotency (id-keyed `IGNORE` on the DAO from PR A4).
  3. Flip to `SENT` / `FAILED` on completion.

## `ChatThreadViewModel`

- Added `retryMessage` closure (defaults to no-op for tests that don't exercise it). Public `retry(messageId)` fires it from `viewModelScope`; errors are swallowed because retry surfaces failures as `MessageStatus.FAILED` on the row, not via thrown exceptions (same posture as the iOS twin).
- `OnymApplication.makeChatThreadViewModel` wires `retryMessage = sender::retry` — same `SendMessageInteractor` instance used for the send path.

## Persistence

Added `findById(id: UUID)` to `MessageStore` + `RoomMessageStore` + `InMemoryMessageStore` + `MessageRepository`. The retry path needs store access without depending on the chat thread being subscribed (the cached per-group flow might not be hot if the user just navigated back).

## Divergence from iOS

- **iOS needed `snapshot.reconfigureItems(...)`** to close the gap where their diffable data source's identity was just `UUID` and a status flip wouldn't repaint. **Compose's recomposition over the `ChatMessage` data class already handles this** — `LazyColumn(key = { it.id })` keeps the slot stable; the bubble body reads the latest fields. PR A6's docstring already called this out.
- **iOS uses a DEBUG-only `simulateBubbleTapForTest()` test seam** because `UITapGestureRecognizer` has no clean public "fire" API. Compose's `Modifier.clickable` is just a callback — no test seam needed for the analogue.

## Tests (16 new)

**`ChatBubbleStatusTest`** (9):
- `statusVisualFor`: pending → muted clock, sent → muted check, failed → error bang, received → null (defensive).
- `canRetry`: failed-outgoing `true`; pending-outgoing, sent-outgoing, failed-incoming all `false` (each gate has its own test).

**`SendMessageInteractorTest`** +6:
- retry failed → `SENT` on relay acceptance.
- retry failed → stays `FAILED` when relays reject.
- retry unknown `messageId` → no-op.
- retry `SENT` → no-op (prevents double-delivery).
- retry `PENDING` → no-op (prevents double-delivery on in-flight).
- retry `INCOMING` → no-op.

**`ChatThreadViewModelTest`** +2:
- retry delegates to interactor closure with `(groupId, messageId)`.
- retry swallows interactor errors.

Full `:app:testDebugUnitTest` suite — **529 / 529 pass**, 0 failures, 0 errors, 0 regressions. `assembleDebug` clean.

## Plan context

PR 9 of N in the Android chat stack. **PR A1** (#141) payload • **PR A2** (#142) persistence • **PR A3** (#143) Ed25519 sender pubkey • **PR A4** (#144) dispatcher + send interactor • **PR A5** (#145) Compose chat-thread shell • **PR A6** (#146) message list rendering • **PR A7** (#147) input panel + keyboard avoidance • **PR A8** (#148) wire Send button to viewModel.send.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/app/onym/android/OnymApplication.kt
@@ -586,6 +586,7 @@ class OnymApplication : Application() {
                     groupRepository = groupRepository,
                     messageRepository = messageRepository,
                     sendMessage = sender::send,
+                    retryMessage = sender::retry,
                 )
             },
             makeIdentitiesViewModel = {
--- a/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
@@ -1,20 +1,30 @@
 package app.onym.android.chats
 
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxWithConstraints
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Check
+import androidx.compose.material.icons.filled.ErrorOutline
+import androidx.compose.material.icons.outlined.Schedule
+import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.dp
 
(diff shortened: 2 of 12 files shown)


dev-onym (commented):
COMMENT — solid PR; correctness looks right and tests cover the gates well, but a few polish items:

- `ChatBubble.kt:159-161` icon modifier order — `.size(STATUS_ICON_SIZE).padding(end = 4.dp)` constrains to 14dp then *insets* by 4dp, leaving a 10dp-wide drawable area (icon scales to ~10x10, not 14x14). For trailing spacing, swap to `.padding(end = 4.dp).size(STATUS_ICON_SIZE)`, or move the spacing onto the parent `Column`.
- `ChatBubble.kt:127` clickable lacks accessibility metadata — no `Role.Button` / `onClickLabel`. The Failed icon's contentDescription is read, but the tap target itself doesn't announce its action. Add `clickable(onClick = onClick, onClickLabel = "Retry")` (or pass `Role.Button` via `semantics`).
- `ChatThreadScreen.kt:194-197` allocates a fresh `{ onRetry(message.id) }` lambda per item per recomposition, which makes `ChatBubble`'s `onRetry` parameter unstable and undercuts LazyColumn skipping. Either hoist with `remember(message.id)` or change `ChatBubble` to take `(UUID) -> Unit` and pass `message.id` separately.
- `SendMessageInteractor.kt:152-173` the FAILED→PENDING flip is read-then-write, not compare-and-set; two near-simultaneous retry taps (UI race vs the PENDING flip) could both pass the gate and double-ship. Practically harmless because (a) UI `canRetry` gates on FAILED and (b) receiver-side `IGNORE` dedups — but the docstring at line 121 ("closes the double-delivery hole") overclaims; the receiver IGNORE is what actually closes it. Either tighten the wording or add a CAS-style status update.
- `SendMessageInteractorTest.kt` retry tests verify `sends().size` but not that the sealed payload carried the *original* `messageId` / `sentAtMillis` — that invariant is the whole basis for receiver dedup and deserves an explicit assertion (decode the recorded send and check the payload fields, the way the send-path tests presumably do).
- Minor: `java.util.UUID` is fully qualified inline in `ChatThreadViewModel.kt:35`, `ChatBubble.kt:122,149`, `ChatThreadScreen.kt:120` — just import it.
- `statusVisualFor(RECEIVED) == null` is dead at the call site (already gated by `isOutgoing`), but the defensive null is cheap and documented — fine to keep.

PR #155, opened 2026-05-22
feat(inbox): create-time invite-offer handshake + verify-at-current group state

lead-onym:
Ports two stacked iOS PRs to Android, mapping each piece to its Android analogue (manual DI, kotlinx.serialization, Compose, coroutines) rather than transliterating Swift:

- onymchat/onym-ios#158 — the invite-offer handshake
- onymchat/onym-ios#159 — verify-at-current group state (supersedes #158's interim staleness window; this PR builds the #159 model, not the window)

## Problem
Create-time "add member via QR" shipped a `GroupInvitationPayload` (a membership grant pinned to one on-chain epoch) directly to a non-member's inbox. The invitee was never added to the on-chain roster nor recorded in the inviter's member directory — **inviter saw 1 member, invitee saw 2, messages didn't flow back** — and the epoch-pinned snapshot went stale the moment any other invitee anchored.

## Handshake (#158)
Route create-time invites into the existing on-chain join machinery (intro keys → `JoinRequestSender` → `IntroInboxPump` → `JoinRequestApprover` → `update_commitment` → member announcement):
- Tyranny create mints a per-invitee intro key and sends a durable, epoch-free `GroupInviteOfferPayload` (grants nothing, never expires) instead of a snapshot. OneOnOne/Anarchy keep the membership-grant send.
- Inbound offers route to a `PendingInvitesStore` and **never materialize a group**.
- The invitee explicitly **Accepts** → ships a `JoinRequestPayload` to the intro key (reuses `JoinRequestSender`).
- The admin explicitly **Approves** via the existing `ApproveRequests` UI (unchanged) — the only on-chain `update_commitment`. No auto-anchor, no silent add.

## Verify-at-current (#159)
A Tyranny snapshot materializes **only** when it byte-verifies against the chain at an **exact epoch** (reproducing `Poseidon(Poseidon(root, epoch), salt)` needs the random salt, which is never on chain — the forgery gate). When the chain has advanced past the snapshot:
- the receiver refuses to trust it and sends a `GroupStateRefreshRequest` to the admin, marking the group **verifying** and keeping it **hidden from the chats list**;
- the admin replies with the current snapshot **only to a current member** whose envelope-signer matches their stored Ed25519, sealing to the member's **stored** inbox (a forged request can't redirect the salt to a non-member);
- on timeout / unknown admin inbox the group is marked **"couldn't verify"** with a **Retry** — never silently materialized or dropped.

The `MemberAnnouncement` verifier keeps its converge-forward (`>=`) gate — it's admin-Ed25519-signed and out of scope for #159.

## Files
**New:** `GroupInviteOfferPayload`, `GroupStateRefreshRequest`, `PendingInvitesStore`, `PendingVerificationStore`, `PendingInvitesViewModel`, `PendingInvitesScreen`, `GroupStateVerifier`.
**Edited:** `CreateGroupInteractor` (Tyranny sends offers), `IncomingMessageDispatcher` (offer + refresh fast-paths; exact-epoch verify with defer-on-stale), `ChatsScreen`/`RootScreen` (Invitations toolbar badge + route), `OnymApplication`/`AppDependencies` (DI wiring + per-identity lifecycle).

## Hard invariants verified
- Roster stays on chain (admin anonymous — never disclosed on chain).
- The only on-chain anchor is the admin's explicit Approve; no auto-anchor, no silent add.
- The offer grants nothing and never expires.
- Salt is disclosed only to a verified current member, sealed to their stored inbox.
- `JoinRequestApprover.approve` was already serialized under a `Mutex` — no change needed.

## Tests
Offer + refresh-request codecs and wire-disambiguation; dispatcher queues offers (no materialize) / routes refresh requests / defers stale snapshots (androidTest FFI: exact-epoch materializes, chain-ahead defers); verifier marks unreachable on unknown admin inbox and refuses non-members / signer-mismatch (salt non-disclosure). **554 unit tests green** (`testDebugUnitTest`); main + androidTest compile.

> Note: instrumentation (androidTest) suites compile but were not executed here (no device/emulator).

### DIFF
--- a/app/src/androidTest/kotlin/app/onym/android/group/CreateGroupInteractorTest.kt
+++ b/app/src/androidTest/kotlin/app/onym/android/group/CreateGroupInteractorTest.kt
@@ -22,6 +22,7 @@ import app.onym.android.support.FakeContractsManifestFetcher
 import app.onym.android.support.FakeKnownRelayersFetcher
 import app.onym.android.support.InMemoryAnchorSelectionStore
 import app.onym.android.support.InMemoryGroupStore
+import app.onym.android.support.InMemoryIntroKeyStore
 import app.onym.android.support.InMemoryRelayerSelectionStore
 import app.onym.android.support.StubGroupProofGenerator
 import kotlinx.coroutines.runBlocking
@@ -219,6 +220,7 @@ class CreateGroupInteractorTest {
             ),
             proofGenerator = StubGroupProofGenerator(),
             inboxTransport = inboxTransport,
+            introducer = InviteIntroducer(InMemoryIntroKeyStore()),
             makeContractTransport = { contractTransport },
         )
 
@@ -251,6 +253,7 @@ class CreateGroupInteractorTest {
             ),
             proofGenerator = StubGroupProofGenerator(),
             inboxTransport = inboxTransport,
+            introducer = InviteIntroducer(InMemoryIntroKeyStore()),
             makeContractTransport = { contractTransport },
         )
 
@@ -529,6 +532,7 @@ class CreateGroupInteractorTest {
         networkPreference = networkPreference,
         proofGenerator = StubGroupProofGenerator(),
         inboxTransport = inboxTransport,
+        introducer = InviteIntroducer(InMemoryIntroKeyStore()),
         makeContractTransport = { contractTransport },
     )
 
--- a/app/src/androidTest/kotlin/app/onym/android/inbox/IncomingMessageDispatcherConvergeForwardFfiTest.kt
+++ b/app/src/androidTest/kotlin/app/onym/android/inbox/IncomingMessageDispatcherConvergeForwardFfiTest.kt
@@ -1,0 +1,179 @@
+package app.onym.android.inbox
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import app.onym.android.chain.ChainStateReading
+import app.onym.android.chain.SepCommitmentEntry
+import app.onym.android.chain.SepGroupType
+import app.onym.android.chain.SepTier
+import app.onym.android.group.GovernanceMember
+import app.onym.android.group.GroupCommitmentBuilder
+import app.onym.android.group.GroupInvitationPayload
+import app.onym.android.group.GroupRepository
+import app.onym.android.identity.IdentityId
+import app.onym.android.persistence.InMemoryInvitationStore
+import app.onym.android.support.FakeActiveIdentityProvider
+import app.onym.android.support.FakeInvitationEnvelopeDecrypter
+import app.onym.android.support.InMemoryGroupStore
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.runBlocking
+import kotlinx.serialization.json.Json
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.time.Instant
+
+/**
+ * FFI-backed verify-at-current tests for [IncomingMessageDispatcher]
+ * (PR 159). Lives in `androidTest/` because the Tyranny invitation
+ * verifier recomputes the Poseidon commitment via [GroupCommitmentBuilder],
+ * whose native `.so` only loads on the device runtime.
+ *
+ *  - exact epoch + matching commitment → materializes (the salt-gated
+ *    forgery gate).
+ *  - chain ahead of the snapshot → deferred to the verifier, NOT
+ *    materialized (an unverifiable snapshot must not become a chat).
+ *
+ * Mirrors `test_invitation_tyranny_chainAhead_defersAndDoesNotMaterialize`
+ * from onym-ios PR #159.
+ */
+@RunWith(AndroidJUnit4::class)
+class IncomingMessageDispatcherConvergeForwardFfiTest {
+
+    private val owner = IdentityId("owner")
+    private val groupId = ByteArray(32) { 0x42 }
+    private val salt = ByteArray(32) { 0x66 }
+
+    @Test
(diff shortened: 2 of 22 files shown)


ux-onym (commented):
APPROVE — strong PR; hard invariants (no auto-anchor, salt only to verified current member, no silent materialization of stale snapshots) hold up under review. Notes below.

- `app/src/main/kotlin/app/onym/android/inbox/GroupStateVerifier.kt:169` — `sendRefresh` builds the request from `activeIdentityId.value` rather than the deferred `target.ownerIdentityId`. The comment acknowledges the single-active V1 assumption, but if the user switches identities between defer and send, the admin's membership check fails and the group is stuck in UNREACHABLE for the wrong reason. Safer to look up `identities.value.firstOrNull { it.id == target.ownerIdentityId }`.
- `app/src/main/kotlin/app/onym/android/inbox/PendingInvitesStore.kt:103` — `consumeForMaterializedGroups` doesn't scope by owner, so a multi-identity user could in theory wipe a pending invite from identity A when identity B materializes the same group_id. Practically impossible (group IDs are 32-byte random), but filtering by `ownerIdentityId` makes the contract obvious.
- `app/src/main/kotlin/app/onym/android/inbox/PendingInvitesStore.kt:67` — `consume(id)` only drops from process memory; the retained Nostr offer re-delivers and re-appears on next launch. Documented as intentional, but worth surfacing in the UI (`Dismiss` reads as durable).
- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:125` — Trial-decode order is safe: offer + refresh both require keys (`intro_pub` / `inviter_alias` / `refresh_group_id` / `requester_inbox_pub`) that no other payload type carries, so a legitimate invitation/announcement cannot be mistakenly queued as an offer. Confirmed by the codec disambiguation tests.
- `app/src/main/kotlin/app/onym/android/inbox/GroupStateVerifier.kt:246-247` — Salt-disclosure gate is correctly layered: BLS-in-roster lookup → Ed25519-signer-matches-stored-sendingPubkey → sealed to the *stored* inbox (not the request's claimed one). Forged-request → salt-redirect is closed.
- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:551-572` — verify-at-current correctly distinguishes chain-behind (REJECT) / chain-equal (byte-verify) / chain-ahead (STALE_NEEDS_REFRESH); FFI test in `IncomingMessageDispatcherConvergeForwardFfiTest.kt` covers both branches.
- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:611-620` — MemberAnnouncement gate relaxed to `>=` (converge-forward) as described. Test pair (`convergeForward_appliedWhenChainEpochAheadOfClaim` + `droppedWhenChainEpochBehindClaim`) covers both directions.
- `app/src/main/kotlin/app/onym/android/group/CreateGroupInteractor.kt:367` — Tyranny send loop correctly mints a *per-invitee* intro key via the shared `InviteIntroducer` so the admin's `IntroInboxPump` picks them up; OneOnOne/Anarchy keep the membership-grant path. Send failures still throw `InvitationSendFailed` after the group is durable (retry-invites UI hook preserved).

dev-onym (commented):
APPROVE

- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:118-145` Trial-decode order is sound: offer requires `intro_pub` + `inviter_alias` (no defaults), refresh requires `refresh_group_id` + `requester_inbox_pub` + `requester_bls_pub` — disjoint from each other and from `GroupInvitationPayload` / `JoinRequestPayload`, so permissive decode can't conflate them.
- `app/src/main/kotlin/app/onym/android/inbox/GroupStateVerifier.kt:209-242` Salt-disclosure gates are correct: BLS membership ∧ envelope-signer == stored Ed25519 ∧ claimed inbox == stored inbox; reply is sealed to `profile.inboxPublicKey` (stored), not the requester's claimed key — forged request can't redirect the salt.
- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:540-572` Converge-forward semantics line up with the PR description: `chain.epoch < claim` → REJECT, `==` byte-verify, `>` → STALE_NEEDS_REFRESH; the `MemberAnnouncement` mirror at :613-624 keeps `>=` gating as advertised.
- `app/src/main/kotlin/app/onym/android/OnymApplication.kt:456,509` Multi-slot removal listeners append (per `IdentityRepository.kt:111`), so the two new registrations don't clobber `GroupRepository` / `IncomingInvitationsRepository` / `MessageRepository` / `introKeyStore` cascades.
- `app/src/main/kotlin/app/onym/android/inbox/PendingInvitesScreen.kt:103-118` Minor: `LazyColumn` keys come from two namespaces (`PendingInvite.id` = Nostr event id, `PendingGroupVerification.id` = groupIdHex, both 64-char lowercase hex). Compose requires globally unique keys across `items()` calls — extremely low collision probability, but consider prefixing (`"invite-$id"` / `"verify-$id"`) to make it impossible.
- `app/src/main/kotlin/app/onym/android/group/GroupInviteOfferPayload.kt:36-39` Doc inaccuracy: comment claims `offer_version` is "required (no defaults)" and contributes to disambiguation, but the field has `val version: Int = 1` so it's optional in decode. Disambiguation actually rests on `intro_pub` + `inviter_alias`; both are truly required, so the guarantee holds — just update the comment.
- `app/src/main/kotlin/app/onym/android/inbox/GroupStateVerifier.kt:124-137` Tiny race: between `delay()` completing and `markUnreachableIfStillVerifying` running, `resolve()` could cancel the timeout job and remove the target while the store still briefly contains the entry, producing a momentary UNREACHABLE flicker before `store.resolveMaterialized` clears it. Cosmetic only — final state is correct.
- `app/src/main/kotlin/app/onym/android/inbox/PendingInvitesViewModel.kt:113-129` `_requestedIds` is in-memory only; on relaunch a sent-but-unapproved invite renders "Accept" again, so user may double-send. Admin's `JoinRequestApprover` should handle idempotency, but worth confirming or adding a UI note.

PR #157, opened 2026-05-23
fix(chats): keep latest messages visible above keyboard (#154)

### LINKED ISSUE #154: Keyboard covers last messages despite input field shifting up
## Summary

In the chat thread view, when the user taps the message input field to bring up the keyboard, the input panel shifts up as expected. However, the last 2–3 single-line messages remain **hidden behind the keyboard** instead of remaining visible above it. An unnecessary vertical gap exists between the keyboard and the input field, which increases the amount of chat history covered by the keyboard.

**Expected behavior:** The message list should scroll so that the last message remains visible just above the input panel (not hidden behind the keyboard). The keyboard should not cover recent messages.

---

## Steps to Reproduce

1. Open any chat that contains multiple messages (at least 4–5 messages).
2. Scroll to the bottom of the chat so the latest messages are visible.
3. Tap on the message input field to focus it and bring up the keyboard.
4. Observe the keyboard opening animation.
5. Observe that the input field shifts up (so it is above the keyboard).
6. Observe that the last 2–3 messages are **still covered by the keyboard**.
7. Note the visible empty gap between the keyboard and the input field.

---

## Expected Behavior

- The keyboard should push the input field up naturally.
- The message list should scroll automatically so that the **most recent message** is visible just above the input field (not hidden).
- No messages should remain covered by the keyboard after focus.
- The gap between the keyboard and input field should be minimal (0–8 dp), not large enough to push extra messages out of view.

---

## Actual Behavior

- Last 2–3 messages remain hidden behind the keyboard.
- An unnecessary vertical gap exists between the keyboard and the input field.
- The user must manually scroll to read messages while typing.

---

## Impact

- **Severity:** Low (no crash, but poor UX)
- **Priority:** Medium
- **User frustration:** Users cannot see recent messages while typing, forcing manual scrolling. This disrupts the natural chat flow.

---

## Possible Root Causes

- **`android:windowSoftInputMode` setting:** May be set to `adjustPan` instead of `adjustResize`, or both modes may not be correctly handling the chat list scrolling.
- **Excess bottom padding:** The input field may have large bottom margins that push the keyboard further down while keeping the list scrolled higher.
- **List scrolling logic:** The RecyclerView (or equivalent) may not be automatically scrolling to the last position when the keyboard opens.
- **Keyboard height detection:** The app may not be correctly accounting for the keyboard's actual height.

---

## Recommendations

1. **Set appropriate soft input mode** in AndroidManifest.xml for the chat activity:
   - Use `android:windowSoftInputMode="adjustResize"` to resize the layout, or
   - `adjustPan` combined with manual scroll logic.

2. **Remove excess gap** between keyboard and input field:
   - Review the chat screen layout XML.
   - Reduce or remove unnecessary bottom margins/padding on the input container.

3. **Auto-scroll to bottom** when keyboard opens:
   - Add a listener for keyboard visibility changes (e.g., using `ViewTreeObserver` or `OnGlobalLayoutListener`).
   - When keyboard becomes visible, scroll the message list to the last position after a short delay.

4. **Test with different keyboard heights** (Gboard, Samsung Keyboard, Yandex Keyboard) to ensure consistent behavior.

lead-onym:
Fixes #154.

When focusing the chat input, the last 2–3 messages stayed hidden behind the keyboard and an unnecessary gap opened between the keyboard and the input panel. Two root causes:

## 1. The gap (double-counted nav-bar inset)

`Scaffold` already feeds the body a `padding` that includes the bottom **navigation-bar** inset, and the body then added `.imePadding()`. The IME inset is measured from the screen bottom and overlaps that same nav-bar band, so the nav-bar height was counted twice — a gap ≈ nav-bar height.

**Fix:** `.consumeWindowInsets(padding)` between `.padding(padding)` and `.imePadding()`, so `imePadding()` contributes only the keyboard height beyond the already-applied inset.

## 2. Hidden messages (LazyColumn pins its top on resize)

`.imePadding()` shrinks the body `Column` when the keyboard rises. A `LazyColumn` keeps its **top** anchored on resize, so the bottom messages slide out of view behind the input panel. The existing auto-scroll only fired on message-count changes, never on keyboard open.

**Fix:** a `LaunchedEffect(WindowInsets.isImeVisible)` re-scrolls to the latest message when the IME appears, gated by the new pure helper `shouldAnchorBottomOnImeShow(imeVisible, nearBottom, hasMessages)`. The `nearBottom` gate reuses the existing append heuristic so focusing the input while reading older history doesn't yank the list down.

## Tests

Added 4 cases to `ChatThreadAutoScrollTest` for the helper (anchors when visible + near-bottom; no-op when IME hidden, scrolled-up, or empty thread). `:app:testDebugUnitTest` passes.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/RootScreen.kt
+++ b/app/src/main/kotlin/app/onym/android/RootScreen.kt
@@ -1,5 +1,6 @@
 package app.onym.android
 
+import androidx.compose.foundation.layout.consumeWindowInsets
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Forum
@@ -151,7 +152,23 @@ fun RootScreen(
         NavHost(
             navController = navController,
             startDestination = Tab.Chats.route,
-            modifier = androidx.compose.ui.Modifier.padding(padding),
+            // `consumeWindowInsets(padding)` is load-bearing, not
+            // decorative: several destinations host their own
+            // `Scaffold` (ChatThreadScreen, SettingsScreen, …). The
+            // outer Scaffold here applies the system-bar insets and
+            // hands them back as `padding`, but does NOT consume them
+            // from the inset tree. Without the consume, a nested
+            // Scaffold reads the full `systemBars` again and re-pads
+            // the bottom navigation-bar inset a second time. On a
+            // scrolling list that doubled inset just looks like extra
+            // bottom padding; on the chat thread's bottom-pinned
+            // input panel it's the visible gap + keyboard-overlap in
+            // #154. Consuming here makes the inner Scaffolds see the
+            // already-applied insets as spent, so `imePadding()`
+            // downstream lands the panel flush on the keyboard.
+            modifier = androidx.compose.ui.Modifier
+                .padding(padding)
+                .consumeWindowInsets(padding),
         ) {
             composable(Tab.Chats.route) {
                 val vm: ChatsViewModel = viewModel(
--- a/app/src/main/kotlin/app/onym/android/chats/ChatThreadScreen.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatThreadScreen.kt
@@ -190,14 +190,16 @@ private fun ChatThreadBody(
         modifier = Modifier
             .fillMaxSize()
             .padding(padding)
-            // The Scaffold's `padding` already carries the bottom
-            // navigation-bar inset. The IME inset is measured from
-            // the screen bottom and overlaps that same nav-bar band,
-            // so applying `imePadding()` on top would count the
-            // nav-bar height twice — the visible gap between the
-            // keyboard and the input panel in #154. Consuming the
-            // scaffold insets first makes `imePadding()` contribute
-            // only the keyboard height beyond what's already padded.
+            // Consume whatever bottom inset this screen's own Scaffold
+            // already applied before `imePadding()` reads the IME
+            // inset. The IME inset is measured from the screen bottom
+            // and overlaps the navigation-bar band, so without the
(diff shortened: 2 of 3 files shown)


dev-onym (commented):
APPROVE

- `ChatThreadScreen.kt:19` — `ExperimentalLayoutApi` import is grouped with `material.icons`/`material3` imports instead of the other `foundation.layout.*` imports at lines 3–12. Pure ordering nit, but the surrounding alphabetical grouping is otherwise clean.
- `ChatThreadScreen.kt:177-187` — `LaunchedEffect(imeVisible)` reads `nearBottom` when the coroutine runs, not when it's launched. The doc-comment's claim that this captures the pre-resize position relies on the IME inset animating in *after* the launcher dispatches; that ordering generally holds but is implicit timing. Capturing `nearBottom` into a local val just before the `LaunchedEffect` (or via `snapshotFlow { imeVisible }.collect { … }` driven from a precomputed remembered state) would make the intent explicit and immune to dispatcher reordering. Not a blocker — current behavior likely works for the bug's reproduction frame budget.
- `ChatThreadScreen.kt:178` — effect also fires on initial composition with `imeVisible=false`; the helper short-circuits so no spurious scroll. Good.
- `ChatThreadScreen.kt:201-202` — `consumeWindowInsets(padding).imePadding()` is the correct ordering to avoid the double-counted nav-bar band; consistent with Compose docs.
- Tests: 4 new cases at `ChatThreadAutoScrollTest.kt:84-132` cover the truth table of the pure helper (the 3-bit input space minus the redundant rows). The doubly-redundant `imeHidden + scrolledUp` row is fine to omit since each false-axis is independently asserted.
- Suggest a follow-up: an instrumentation/UI test would catch any regression in the IME→layout→coroutine ordering the helper depends on, since the pure-helper tests can't.

Issue #88, opened 2026-05-05
Cannot add participants via QR or invite link during group creation


alexpovstin:
## Summary

When creating a group on smartphone **A**, the participant addition step presents a field for entering a 64-character key. However:

1. There is no obvious way to obtain such a 64-character key from another user.
2. There is **no QR scan button** on this screen.

This makes it impossible to add a participant from smartphone **M** during group creation.

---

## Steps to Reproduce

1. On smartphone **A**, start creating a new group (e.g., "Tyranny").
2. Proceed to the **"Add participants"** step.
3. Observe a text input field labeled for a 64-character key.
4. On smartphone **M**, locate the user's invite QR code or invite link (e.g., `https://onym.chat/i?k=SDhSwaJku0X1yofBvvb5suEiBaz5euq-r3Q8qhNoE8`).
5. Attempt to add the user from **M** to the group on **A**:
   - **Option A (QR):** There is no QR scan button on **A** to scan **M**'s QR code.
   - **Option B (link/key):** Pasting the full invite link into the 64-character key field does not work – the app does not recognize or parse the link.
6. On smartphone **A**, try opening the invite link directly (e.g., in a browser). The link opens the promotional website `onym.chat` instead of triggering any in-app action.

---

## Expected Behavior

The user should be able to add a participant during group creation using one of the following methods:

- **Scan QR code** – a visible QR scan button on the "Add participants" screen.
- **Paste invite link** – the app should automatically recognize and extract the 64-character key from a full invite link (e.g., `https://onym.chat/i?k=...`).
- **Copy 64-character key from identity settings** – users should be able to copy their own invite key from within the app to share with others.

Additionally, clicking an invite link (`https://onym.chat/i?k=...`) on an Android device with the app installed should open the app and trigger the add/join flow (deep linking).

---

## Actual Behavior

- No QR scan button on the group creation participant step.
- The 64-character key field accepts only raw keys, not full invite links.
- Invite links open the promotional website instead of the app.
- Users cannot copy their own 64-character invite key from anywhere in the app settings.

---

## Impact

- **Severity:** Medium (blocks adding participants during group creation)
- **Priority:** High (core collaboration feature is broken)
- **User frustration:** Users are forced to create the group first, then add members later via group info (workaround exists but is unintuitive).

---

## Proposed Solutions

1. **Add a QR scan button** on the "Add participants" screen during group creation.
2. **Add an option to copy the 64-character invite key** somewhere in Identity Settings, so users can share it manually.
[...]

PR #159, opened 2026-05-23
feat(groups): scan-QR invite on create-group flow (#88)

lead-onym:
## What

Adds a **Scan QR code** affordance to the Create Group → *Invite by inbox key* screen, mirroring onym-ios. The user scans a contact's identity-invite QR (`https://onym.app/i?k=…`); the decoded string is canonicalised to the bare 64-char hex inbox key and dropped into the existing field for review before *Add invitee*. A malformed scan surfaces the same inline error as a malformed paste.

Closes part of #88 (the Scan-QR ask — see Scope below).

## How

| Piece | Detail |
|---|---|
| `group/InviteKeyCanonicalizer.kt` | Pure-JVM twin of iOS `canonicalizeInviteKey(_:)`. Recognises bare hex, `/i?k=<base64url>` (identity invite), and legacy iOS `?payload=<hex>`. Falls through to trimmed input so the caller's hex validation produces a meaningful error. Unit-tested, no Robolectric. |
| `scan/QrCodeAnalyzer.kt` | CameraX `ImageAnalysis.Analyzer` decoding the Y-plane with zxing-core `MultiFormatReader` (QR-only, `TRY_HARDER`, inverted fallback). |
| `scan/QrScannerScreen.kt` | Full-screen CameraX preview, runtime `CAMERA` permission, viewfinder + cancel chrome matching the iOS `QRCodeScannerView`, one-shot dedup latch. |
| `CreateGroupViewModel.tappedScannedKey` | Canonicalises into `inviteeInput`; validation stays in `tappedAddInvitee`. |
| Plumbing | `CAMERA` permission + `uses-feature camera.any required="false"`; CameraX 1.4.1 in the version catalog; en + ru strings. |

### Dependency choice
Camera layer is **CameraX**; decode reuses the **zxing-core** dependency already in the APK (encode-only until now). No ML Kit model and no Google Play Services — keeps the APK **F-Droid-buildable** (both deps are Apache-2.0, source-buildable).

## Scope

Scan path only. The other #88 asks are deliberate follow-ups:
- pasting a full invite link into the key field,
- copying your own inbox key from Identity Settings,
- `onym.app/i` deep linking (the manifest deep-links `/join` for group invites but not `/i` for identity invites).

## Testing

- `InviteKeyCanonicalizerTest` — 8/8 pass (bare hex case-preserved, `/i?k=`, custom-scheme `k=`, legacy `?payload=`, unknown/garbled fallbacks).
- `:app:compileDebugKotlin` and `:app:lintDebug` (incl. hard-gated `MissingTranslation`) pass.
- ⚠️ Live camera decode not exercised on hardware — worth a device smoke test (CameraX Y-plane → zxing is the standard path, but verify against a real `/i?k=` QR).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -163,6 +163,15 @@ dependencies {
     // renderer).
     implementation(libs.zxing.core)
 
+    // CameraX — preview + image analysis behind the invite-QR scanner
+    // (`QrScannerScreen`). camera-view supplies `PreviewView`; the
+    // analysis frames are decoded with the zxing-core reader above so
+    // we add no ML Kit model to the APK.
+    implementation(libs.androidx.camera.core)
+    implementation(libs.androidx.camera.camera2)
+    implementation(libs.androidx.camera.lifecycle)
+    implementation(libs.androidx.camera.view)
+
     implementation(libs.onym.sdk)
 
     // Room — `suspend` DAO + KSP-generated bindings. PersistenceStore
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,23 @@
     -->
     <uses-permission android:name="android.permission.INTERNET" />
 
+    <!--
+        CAMERA: the invite-QR scanner on the Create Group → "Invite by
+        inbox key" screen (`QrScannerScreen`) reads a contact's
+        `https://onym.app/i?k=…` identity-invite QR via CameraX.
+        Requested at runtime (API 23+) the first time the user taps
+        "Scan QR"; denial degrades to the existing paste-the-key path.
+
+        `uses-feature ... required="false"`: a camera is not essential
+        to the app — devices without one (or where the user declines)
+        still use manual key entry — so we don't want Play Store to
+        filter the listing on camera hardware.
+    -->
+    <uses-permission android:name="android.permission.CAMERA" />
+    <uses-feature
+        android:name="android.hardware.camera.any"
+        android:required="false" />
+
     <application
         android:name=".OnymApplication"
         android:label="@string/app_name"
--- a/app/src/main/kotlin/app/onym/android/group/CreateGroupViewModel.kt
+++ b/app/src/main/kotlin/app/onym/android/group/CreateGroupViewModel.kt
@@ -293,6 +293,22 @@ class CreateGroupViewModel(
         )
     }
 
+    /**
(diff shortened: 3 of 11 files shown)


dev-onym (commented):
COMMENT — solid mirror of the iOS flow, tests cover the canonicalizer well, and CameraX/zxing wiring looks correct. A few worth-considering items:

- `QrScannerScreen.kt:80-120` — no `BackHandler { onCancel() }`; hardware Back falls through to the nav backstack and pops `InviteByKey` instead of just dismissing the overlay. Likely a UX regression vs. tapping the X chrome.
- `QrScannerScreen.kt:101-104` — on permanent denial the `LaunchedEffect(Unit)` only fires once, so the user is stranded on the "permission denied" panel with no path forward. Consider an "Open Settings" affordance (intent to `ACTION_APPLICATION_DETAILS_SETTINGS`) since the copy already tells them to enable it in Settings.
- `QrScannerScreen.kt:174-189` — when the user grants permission *after* the initial launcher result, `hasPermission` flips and `CameraPreview` enters composition; that's fine. But if the `ProcessCameraProvider.getInstance` future never resolves before `onDispose` (rare, but possible if user cancels fast), `providerFuture.get()` will block the main thread inside onDispose. Consider `if (providerFuture.isDone) providerFuture.get().unbindAll()` to avoid the blocking get.
- `QrCodeAnalyzer.kt:78` — `reader.reset()` between frames is fine for hints (MultiFormatReader keeps them in a separate field), but it discards the per-decoder state that `decodeWithState` is designed to amortise; you might as well call `decode()` instead, or drop the reset and let state accumulate. Minor.
- `InviteKeyCanonicalizer.kt:38-45` — `URI(trimmed).rawQuery` returns the un-percent-decoded value, but `value` is used as-is for Base64 decode. Works today because the emitter uses URL-safe Base64 without padding (no `+`/`/`/`=`), but if a future emitter percent-encodes `=` padding the decode will fail silently and fall through. Add a `URLDecoder.decode(value, UTF_8)` step or document the no-percent-encoding contract.
- `CreateGroupScreen.kt:524-525` — wrapping `Box { Column { … } }` with both at `fillMaxSize()` works but the indentation inside `InviteByKeyScreen` is off (the inner `Column` keeps its old indent), which makes the brace structure hard to follow on review. Cosmetic.
- `AndroidManifest.xml:16-30` — `<uses-feature camera.any required="false">` is correct for the F-Droid story; nice catch.

PR #160, opened 2026-05-23
feat(identity): restore from 12/24-word recovery phrase (#86)

### LINKED ISSUE #86: No way to restore identity using 12-word recovery phrase

lead-onym:
## Summary

Closes #86 — there was no way to restore an existing identity from a 12-word recovery phrase on Android.

This mirrors the iOS feature (`IdentitiesView.AddIdentitySheet`), which puts restore in the **Add Identity** flow rather than the Privacy & Encryption key list. The Identities-screen "Add Identity" button now opens a dialog with:
- an optional **name** field, and
- a **recovery phrase** field.

Leaving the phrase blank mints a fresh BIP-39 identity (the previous one-tap behavior); pasting a 12 or 24-word phrase restores an existing one. Invalid phrases surface an inline error and keep the typed input so the user can fix it.

The repository layer (`IdentityRepository.add(name, restoreMnemonic)`) already supported restore — this PR wires the ViewModel and UI.

## Changes

- **`IdentitiesViewModel`** — `add` now takes `restoreMnemonic`; added an `AddResult` (`Success` / `InvalidMnemonic` / `Failure`) flow so the dialog dismisses on success and inlines errors otherwise. Removed the now-dead `Error.Add` snackbar branch.
- **`IdentitiesScreen`** — new `AddIdentityDialog` (`AlertDialog`, matching the existing `RemoveIdentityDialog` pattern).
- **`strings.xml`** — dialog title, labels, hint, invalid-phrase message.
- **`IdentitiesUITest`** — updated the two tests that assumed one-tap-appends; added valid-phrase restore (asserts persisted entropy round-trips to the canonical `abandon × 11 + about` vector, proving restore vs. fresh mint) and invalid-phrase (asserts inline error + no identity added) cases.

## Testing

- `:app:compileDebugKotlin` and `:app:compileDebugAndroidTestKotlin` pass.
- ⚠️ Instrumented `IdentitiesUITest` cases compile but were **not executed** — no device/emulator was available in this environment. Worth a run on device.

## Scope note

The issue also mentions the non-clickable "Recovery Phrase / Nostr Identifier / …" buttons in Privacy & Encryption. Those are key *display/backup* affordances; per the issue's own "Expected Behavior" (restore "ideally in the Identities section") and the iOS reference, restore lives in Add Identity. Making those Privacy buttons interactive is left as a separate concern.

### DIFF
--- a/app/src/androidTest/kotlin/app/onym/android/uitests/IdentitiesUITest.kt
+++ b/app/src/androidTest/kotlin/app/onym/android/uitests/IdentitiesUITest.kt
@@ -127,14 +127,72 @@ class IdentitiesUITest {
     }
 
     @Test
-    fun addIdentity_appendsRow() {
+    fun addIdentity_blankPhrase_appendsFreshRow() {
+        settings.tapIdentitiesRow()
+        composeRule.waitUntil(timeoutMillis = 10.seconds.inWholeMilliseconds) {
+            identityStore.listIds().size == 1
+        }
+        // "Add Identity" now opens a dialog; leaving the recovery-phrase
+        // field blank mints a fresh identity on confirm.
+        composeRule.onNodeWithTag("identities.add").performClick()
+        composeRule.waitUntil(timeoutMillis = 5.seconds.inWholeMilliseconds) {
+            composeRule.onAllNodesWithTag("identities.add.confirm").fetchSemanticsNodes().isNotEmpty()
+        }
+        composeRule.onNodeWithTag("identities.add.confirm").performClick()
+        composeRule.waitUntil(timeoutMillis = 5.seconds.inWholeMilliseconds) {
+            identityStore.listIds().size == 2
+        }
+    }
+
+    @Test
+    fun addIdentity_validPhrase_restoresDeterministicIdentity() {
         settings.tapIdentitiesRow()
         composeRule.waitUntil(timeoutMillis = 10.seconds.inWholeMilliseconds) {
             identityStore.listIds().size == 1
         }
         composeRule.onNodeWithTag("identities.add").performClick()
         composeRule.waitUntil(timeoutMillis = 5.seconds.inWholeMilliseconds) {
+            composeRule.onAllNodesWithTag("identities.add.mnemonic").fetchSemanticsNodes().isNotEmpty()
+        }
+        composeRule.onNodeWithTag("identities.add.mnemonic").performTextInput(CANONICAL_MNEMONIC)
+        composeRule.onNodeWithTag("identities.add.confirm").performClick()
+
+        composeRule.waitUntil(timeoutMillis = 5.seconds.inWholeMilliseconds) {
             identityStore.listIds().size == 2
+        }
+        // The restored identity becomes active. Its persisted entropy
+        // must round-trip to the phrase we typed — proving restore, not
+        // a fresh mint. (Determinism of the derived keys is covered at
+        // the repository level; here we assert the wiring restores the
+        // exact phrase.)
+        val activeId = identityStore.loadCurrent()!!
+        val snapshot = identityStore.load(activeId)!!
+        val restoredPhrase = app.onym.android.identity.Bip39.mnemonicFromEntropy(snapshot.entropy!!)
+        assert(restoredPhrase == CANONICAL_MNEMONIC) {
+            "expected restored phrase to match input, got: $restoredPhrase"
+        }
+    }
+
+    @Test
+    fun addIdentity_invalidPhrase_showsInlineError_andDoesNotAdd() {
+        settings.tapIdentitiesRow()
+        composeRule.waitUntil(timeoutMillis = 10.seconds.inWholeMilliseconds) {
+            identityStore.listIds().size == 1
+        }
+        composeRule.onNodeWithTag("identities.add").performClick()
+        composeRule.waitUntil(timeoutMillis = 5.seconds.inWholeMilliseconds) {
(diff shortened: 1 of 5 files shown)


dev-onym (commented):
APPROVE — restore wiring is clean and matches the iOS pattern; tests cover blank/valid/invalid paths.

- `app/src/main/res/values-ru/strings.xml`: the 6 new `identities_add_*` keys are not translated; Russian users will see English in the dialog. The existing file already translates `identities_add_button` and `identities_error_add`, so this is a localization regression rather than an unsupported feature.
- `IdentitiesViewModel.kt:101`: `restoreMnemonic?.trim()?.ifBlank { null }` is redundant — `IdentitiesScreen.kt:284` already passes `mnemonic.trim().ifBlank { null }`. Harmless, but pick one layer.
- `IdentitiesScreen.kt:257-260`: mnemonic `KeyboardOptions` sets no `imeAction`, so the IME shows a newline button instead of "Done"; also `autoCorrect = false` is deprecated in current Compose (use `autoCorrectEnabled`). Both minor.
- `IdentitiesScreen.kt:223-229`: `AddResult.Failure` is rendered inline via `identities_error_add` ("Couldn't add identity: %1$s") — that string was previously a Snackbar message, so it reads a bit verbose inside the dialog, but acceptable.
- `IdentitiesScreen.kt:282-289`: `confirmButton` is always enabled; rapid double-taps will enqueue two `identity.add` calls (serialized by the repo mutex) and a fresh-mint double-tap creates two identities. Consider disabling while a submit is in flight. Not a regression vs. the prior one-tap behavior, but the dialog makes it easier to hit.
- `IdentitiesUITest.kt:177`: assertion uses Kotlin `assert(...)` which is a no-op unless `-ea` is set on the JVM running the test — on Android instrumentation this typically doesn't fire. Prefer `assertEquals` / `assertTrue` from `kotlin.test` or JUnit so failures actually fail the test. (Same issue at line 199.)
- PR body flags that instrumented tests weren't executed — worth a device run before merge, especially since the `assert` issue above means the round-trip check could silently no-op.

dev-onym (commented):
APPROVE — solid PR; logic is sound, exceptions wired correctly, strings present in both locales, tests are well-targeted.

- `IdentityError.InvalidMnemonic` exists and is caught explicitly (IdentitiesViewModel.kt:104); other throwables fall through to `Failure`, which the dialog renders via the still-present `identities_error_add` template — verified.
- Minor (secrets): `AddResult.Failure(causeText())` is `e.message ?: simpleName` (IdentitiesViewModel.kt:142, :107) and gets rendered inline via `identities_error_add` (IdentitiesScreen.kt:227). For `IdentityError.InvalidMnemonic` the message is static, but for unknown throwables there's no guarantee the message doesn't echo the typed phrase. Consider suppressing the cause text for the recovery-phrase path or whitelisting known causes.
- Minor (UX/race): the Confirm button isn't disabled while `add` is in flight (IdentitiesScreen.kt:282–288). A fast double-tap launches two `viewModel.add()` coroutines and can append two identities before `LaunchedEffect(addResult)` closes the dialog. Cheap fix: gate `onClick` on `addResult == null` or track an `isAdding` flag.
- Minor (state): dialog uses `remember { mutableStateOf("") }` for both fields (IdentitiesScreen.kt:220–221) — typed phrase is lost on configuration change. For secret material this is arguably desirable; just confirm the trade-off is intentional vs. `rememberSaveable`.
- Tests cover blank/valid/invalid (IdentitiesUITest.kt:131,150,178); using `assertCountEquals(0)` to assert absence of the inline error is the right pattern. PR notes already disclose instrumented tests weren't executed — worth a run on device before merge.
- i18n: en + ru strings added; placeholder match (`%1$s`) on `identities_error_add` lines up (values/strings.xml:473, values-ru/strings.xml:469).

PR #162, opened 2026-05-23
feat(chats): differentiate message senders without avatars

lead-onym:
## Why

Group chats showed **no per-message sender identity** beyond left/right alignment and a fixed color. With no avatars, consecutive incoming messages from different people were indistinguishable.

Ports onym-ios#162 to Android, matching its behavior.

## What

Sender differentiation built on the **load-bearing identity (BLS pubkey)** rather than the self-asserted, spoofable `MemberProfile.alias`:

- **`OnymAccent.forSender(blsPubkeyHex)`** — deterministic accent hashed (FNV-1a) from the BLS pubkey hex. **Byte-identical to iOS**: same constants (offset `0xcbf29ce484222325`, prime `0x100000001b3`), same UTF-8 byte input, same `% 6` into the same palette order (orange, blue, green, purple, pink, yellow). A person reads as the same color on both platforms; the color doubles as a cheap visual fingerprint an alias-impostor can't forge.
- **Incoming runs** get an accent-colored **name header** at the start of each consecutive same-sender run; mid-run messages stay bare. The bubble carries the sender's accent at 20% tint, body text on the normal `onSurface` token for readability. Name = alias, falling back to `BLS xxxxxxxx` (first 8 hex chars) when unset.
- **Outgoing bubbles** fill with the user's **own** hashed accent (own messages already carry `myBlsHex`), so "your color" is consistent with the members list and across groups.
- **1-on-1 groups** (`SepGroupType.ONE_ON_ONE`) suppress the header.
- **Live refresh**: headers repaint as profiles change (joiner admitted, alias edited) without a new message — the screen builds displays off the live `group` snapshot, so Compose recomposes in place (Android's equivalent of iOS's `reconfigureItems`).

## Design

Run-grouping + alias resolution live in the screen layer as a pure `buildSenderDisplays(messages, memberProfiles)`; the `ChatBubble` stays a dumb renderer that receives a resolved `ChatSenderDisplay`. (iOS puts this in the view controller; Android puts it in the screen.)

Note: the Chats tab isn't wrapped in `OnymTheme`, so I added `OnymAccent.color(darkTheme)` to resolve the accent against `isSystemInDarkTheme()` (the no-arg `color()` defaults to the dark variant outside an `OnymTheme` provider).

## Files

- `group/OnymBrand.kt` — `OnymAccent.forSender` + `color(darkTheme)`
- `chats/ChatSenderDisplay.kt` (new) — `ChatSenderDisplay` + pure `buildSenderDisplays` / `senderName`
- `chats/ChatBubble.kt` — accepts `ChatSenderDisplay`, renders name header + per-sender coloring
- `chats/ChatThreadScreen.kt` — builds displays from messages + `group.memberProfiles`, pipes to each bubble

## Tests

- `OnymAccentSenderColorTest` — determinism, palette spread, palette membership, and a reference-FNV-1a vector test pinning cross-platform parity.
- `ChatSenderDisplayTest` — run start / mid-run / sender change, outgoing suppression, 1-on-1 suppression, alias resolution, fingerprint fallback, live profile refresh, and pubkey-keyed (alias-independent) accent.

Full `:app:testDebugUnitTest` suite green; `compileDebugKotlin` + `compileDebugAndroidTestKotlin` clean; secret-read lint clean.

## Note

The BLS fingerprint fallback uses 8 hex chars (`BLS aaaaaaaa`) to match iOS #162 exactly. The existing `ChatMembersScreen` uses a 12-char prefix + ellipsis — a pre-existing minor inconsistency left untouched here.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt
@@ -2,6 +2,7 @@ package app.onym.android.chats
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxWithConstraints
@@ -26,18 +27,32 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 
 /**
  * Chat-thread bubble. Two visual modes driven by [ChatMessage.direction]:
  *
- *  - [MessageDirection.OUTGOING] — `colorScheme.primary` fill,
- *    `colorScheme.onPrimary` text, trailing-aligned. Status glyph
+ *  - [MessageDirection.OUTGOING] — filled with the sender's accent
+ *    ([ChatSenderDisplay.accent], which for own messages hashes to the
+ *    user's own color), `onAccent` text, trailing-aligned. Status glyph
  *    below the trailing edge (clock / check / red bang).
- *  - [MessageDirection.INCOMING] — `colorScheme.surfaceVariant`
- *    fill, `colorScheme.onSurfaceVariant` text, leading-aligned.
- *    No status indicator (the message arriving is proof of
+ *  - [MessageDirection.INCOMING] — tinted with the sender's accent at
+ *    [INCOMING_TINT_ALPHA], `onSurface` text for readability,
+ *    leading-aligned, optionally topped by an accent-colored name
+ *    header. No status indicator (the message arriving is proof of
  *    delivery).
+ *
+ * Sender differentiation (onym-ios PR #162): there are no avatars, so
+ * consecutive incoming messages from one person are grouped under a
+ * single accent-colored name header ([ChatSenderDisplay.showNameHeader],
+ * decided by the screen's run-grouping pass) and the bubble carries
+ * that sender's accent. The accent is a hash of the BLS pubkey, not the
+ * alias, so it doubles as a cheap visual fingerprint an alias-spoofer
+ * can't forge. The bubble stays a dumb renderer — it receives a
+ * resolved [ChatSenderDisplay] and never looks up aliases or hashes
+ * pubkeys itself.
  *
  * Bubble max width is capped at [maxWidthFraction] of the parent
  * row (default 75%) so a long monologue doesn't reach the opposite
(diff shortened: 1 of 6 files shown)


dev-onym (commented):
APPROVE

- `app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt:103` — Reading `isSystemInDarkTheme()` inside the bubble works, but with potentially many bubbles in a `LazyColumn` each one re-subscribes; consider resolving once in `ChatThreadBody` and threading `accentColor` (or the `darkTheme` flag) through `ChatSenderDisplay` / a `CompositionLocal`. Minor.
- `app/src/main/kotlin/app/onym/android/chats/ChatBubble.kt:101-110` — Outgoing text-color rule `if (darkTheme) Black else White` is a bare bool→color flip with only a doc comment as the contract to `OnymAccent` variants. If iOS ever introduces a palette entry whose dark variant is dark, this silently goes unreadable. A small `OnymAccent.onAccent(darkTheme)` (or a token lookup) co-located with the palette would keep the invariant where the colors live. Optional nit.
- `app/src/main/kotlin/app/onym/android/chats/ChatSenderDisplay.kt:43` — `Unknown` default accent is hard-coded `Blue`. Since the screen always populates `senderDisplays`, this is only a defensive default — fine, but the choice of `Blue` (a palette member) means a stray unmapped message would look like a real "Blue sender." A neutral surface-derived color would be safer, though in practice unreachable. Optional.
- `app/src/main/kotlin/app/onym/android/group/OnymBrand.kt:181-188` — FNV-1a constants and `% entries.size` over UTF-8 bytes match the iOS spec; reference-vector test at `app/src/test/kotlin/app/onym/android/group/OnymAccentSenderColorTest.kt:56` pins it. Good cross-platform guard.
- `app/src/main/kotlin/app/onym/android/chats/ChatSenderDisplay.kt:73-78` — Run-grouping correctly treats outgoing messages as breaking the run (previousSender updates unconditionally), so an incoming → outgoing → same-incoming sequence re-shows the header. Matches the documented policy.
- Tests cover determinism, palette spread, palette membership, FNV-1a reference vectors, run start/mid/change, outgoing & 1-on-1 suppression, alias resolution, BLS fingerprint fallback, live profile refresh, and pubkey-keyed accent independence — comprehensive for a pure layer.
- `app/src/main/kotlin/app/onym/android/chats/ChatThreadScreen.kt:143-146` — `remember(sortedMessages, memberProfiles)` correctly recomputes on profile updates; `group?.memberProfiles.orEmpty()` returns the singleton empty map when null, so no spurious invalidation.

ux-onym (commented):
**APPROVE** — clean port with strong test coverage and good separation (pure run-grouping function + dumb renderer).

- `OnymBrand.kt:180-188` — FNV-1a implementation correct; constants and palette order pinned cross-platform and guarded by `OnymAccentSenderColorTest.forSender_matchesReferenceFnv1a`. UTF-8 byte path via `encodeToByteArray()` matches iOS exactly. 
- `ChatBubble.kt:84` — `isSystemInDarkTheme()` is read directly. If a future change wraps the Chats tab in `OnymTheme` with a forced override (e.g. Settings theme toggle), the bubble's accent will diverge from `OnymAccent.color()` callers reading `LocalOnymTokens`. The comment acknowledges this, but consider a single `OnymAccent.colorAuto()` that prefers `LocalOnymTokens` and falls back to system to avoid two source-of-truths.
- `ChatBubble.kt:99-105` — White-on-Yellow (`0xFFD9A400`) outgoing fill in light mode is likely below WCAG AA contrast (~2.8:1). Mirrors iOS so parity-preserving, but worth a known-issue note.
- `ChatBubble.kt:128-139` — Accent-colored name header on `colorScheme.surface` has the same concern; Yellow header text in light mode may be hard to read against white.
- `ChatSenderDisplay.kt:60-72` — `previousSender` tracks across outgoing too, so `A → me → A` re-shows A's header. Probably intended (the outgoing visually breaks the run), but the test suite doesn't cover this specific case — consider adding a `outgoingBetweenIncoming_restartsHeader` assertion to lock the policy.
- `ChatSenderDisplay.kt:85-93` — `senderName` lookup is O(1) but called on every bubble in `buildSenderDisplays`; fine. Note that an empty alias (`""`) falls back to fingerprint — consistent with `ChatMembersScreen`.
- 8-char vs 12-char fingerprint inconsistency with `ChatMembersScreen` (noted in PR description) — acceptable as scoped-out, but file a follow-up so the next reader doesn't have to rediscover this.
- Tests cover determinism, palette membership, spread, FNV-1a reference parity, run-start/mid-run/sender-change, outgoing/1-on-1 suppression, alias resolution, fingerprint fallback, live profile refresh, and pubkey-keyed accent independence. Thorough.

PR #165, opened 2026-05-23
feat(groups): group avatars, wire-compatible with iOS

lead-onym:
Brings Android to parity with the iOS group-avatar stack (`onymchat/onym-ios` #164–#166). An avatar set or changed on either platform renders on the other, including removal; non-admin avatar messages are rejected; photo-less groups are unaffected.

## Wire format / interop
- **`GroupInvitationPayload`** — optional base64 `avatar` field (matches Swift `Data`). "If present" decode: pre-avatar senders omit the key and still decode. Included on send (`JoinRequestApprover`) and applied at materialize time. This is the only path that delivers the photo to **create-time** members (esp. Tyranny, where the full snapshot is sent at join-approval).
- **`GroupAvatarPayload`** (new) — dedicated admin-signed control message with the distinct `avatar_*` keys so it can't be structurally confused with a chat message. Absent `avatar` = removal; `avatar_sender_bls_hex` is informational only (never authenticated with).

## Send / receive
- **`GroupAvatarBroadcaster`** (new) — admin-only. Applies + persists locally first (best-effort broadcast), then seals one envelope per member inbox **except the admin's own key**, on the same X25519 + AES-GCM + Ed25519 path as member announcements.
- **`IncomingMessageDispatcher`** — new decode branch **before** the chat-message branch. Trust gate identical to member announcements (verified Ed25519 signer must equal the stored admin; skipped for admin-less/legacy groups), idempotency no-op, clear-on-absent. **No on-chain commitment check** — the avatar isn't part of the cryptographic group state.

## Persistence
`ChatGroup.avatar` + `PersistedGroup.encryptedAvatar` (AES-GCM at rest), `GroupDatabase` v5→v6 (`MIGRATION_5_6`, non-destructive), wired in `OnymApplication`.

## Image pipeline
`GroupAvatarImage.encode` — centre-crop to square → 256×256 → JPEG quality loop from ~0.8 down until ≤ 16 KB raw (≈22 KB base64), keeping sealed envelopes within relay event-size limits.

## UI
`ChatMembersScreen` shows the group photo (decoded JPEG, or a monogram placeholder). Admins get tap-to-pick + Add/Change/Remove via the system photo picker; non-admins see it read-only.

## Tests
- Invitation payload: round-trips with/without `avatar`; legacy (no key) → null; key omitted when nil.
- `GroupAvatarPayload`: round-trip; absent `avatar` = removal; does **not** decode as a chat message and vice-versa.
- Dispatcher: applied on admin-signer match; rejected on mismatch; cleared on absent; dropped for unknown group; best-effort when no stored admin.
- Broadcaster: fan-out to all-but-admin; non-admin gate; removal omits the key.
- `RoomGroupStore`: encrypted-avatar round-trip + null column.
- Image pipeline (androidTest — real JPEG encoding): 256×256 & ≤ 16 KB on high-entropy and non-square input.

`./gradlew :app:testDebugUnitTest` passes; main + androidTest source sets compile clean. The instrumented image-pipeline test requires a device/emulator to run.

## Note
Production invitation serialization (`encodeDefaults=true`) emits `"avatar":null` when there's no photo — consistent with how `commitment`/`member_profiles` already behave, and iOS `decodeIfPresent` reads it as nil. The type fully supports omission (covered by `encode_omitsAvatarKeyWhenNil`); the actual JPEG bytes are only present when an avatar is set, so photo-less invites stay small.

### DIFF
--- a/app/src/androidTest/kotlin/app/onym/android/group/GroupAvatarImageTest.kt
+++ b/app/src/androidTest/kotlin/app/onym/android/group/GroupAvatarImageTest.kt
@@ -1,0 +1,60 @@
+package app.onym.android.group
+
+import android.graphics.Bitmap
+import android.graphics.Color
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import kotlin.random.Random
+
+/**
+ * Instrumented tests for [GroupAvatarImage]. Lives in `androidTest/`
+ * because real JPEG encoding + bitmap scaling run on the device runtime
+ * (Robolectric's bitmap shadows don't produce real, size-bounded JPEG
+ * bytes, so the 16 KB budget can't be verified off-device).
+ *
+ * Mirrors `GroupAvatarImageTests.swift` from onym-ios PR #164.
+ */
+@RunWith(AndroidJUnit4::class)
+class GroupAvatarImageTest {
+
+    @Test
+    fun encode_highEntropySource_isSquare256AndUnderBudget() {
+        // Worst case for JPEG: a large, full-resolution noise image —
+        // high entropy resists compression, stressing the quality loop.
+        val noise = Bitmap.createBitmap(1024, 768, Bitmap.Config.ARGB_8888)
+        val rng = Random(42)
+        for (y in 0 until noise.height) {
+            for (x in 0 until noise.width) {
+                noise.setPixel(
+                    x,
+                    y,
+                    Color.rgb(rng.nextInt(256), rng.nextInt(256), rng.nextInt(256)),
+                )
+            }
+        }
+
+        val encoded = GroupAvatarImage.encode(noise)
+
+        assertTrue(
+            "encoded JPEG must be within the 16 KB budget, was ${encoded.size}",
+            encoded.size <= GroupAvatarImage.MAX_BYTES,
+        )
+        val decoded = android.graphics.BitmapFactory.decodeByteArray(encoded, 0, encoded.size)
+        assertEquals(GroupAvatarImage.SIZE, decoded.width)
+        assertEquals(GroupAvatarImage.SIZE, decoded.height)
+    }
+
+    @Test
+    fun encode_nonSquareSource_isCentreCroppedToSquare() {
+        val wide = Bitmap.createBitmap(800, 200, Bitmap.Config.ARGB_8888)
+        wide.eraseColor(Color.BLUE)
+        val encoded = GroupAvatarImage.encode(wide)
+        val decoded = android.graphics.BitmapFactory.decodeByteArray(encoded, 0, encoded.size)
+        assertEquals(GroupAvatarImage.SIZE, decoded.width)
+        assertEquals(GroupAvatarImage.SIZE, decoded.height)
+        assertTrue(encoded.size <= GroupAvatarImage.MAX_BYTES)
+    }
+}
--- a/app/src/main/kotlin/app/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/app/onym/android/OnymApplication.kt
@@ -290,6 +290,7 @@ class OnymApplication : Application() {
                 .addMigrations(
                     GroupDatabaseMigrations.MIGRATION_3_4,
                     GroupDatabaseMigrations.MIGRATION_4_5,
+                    GroupDatabaseMigrations.MIGRATION_5_6,
                 )
                 .fallbackToDestructiveMigration()
                 .build()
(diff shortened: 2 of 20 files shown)


ux-onym (commented):
APPROVE

- `app/src/main/kotlin/app/onym/android/group/GroupAvatarBroadcaster.kt:71-77`: returning `Outcome.Sent` when payload serialization throws is misleading — local apply happened but fan-out was skipped. Either propagate a distinct outcome (e.g. `LocallyAppliedNotSent`) or, since the JSON shape is statically known and can't realistically fail to encode, drop the catch and let it surface. Low severity.
- `app/src/main/kotlin/app/onym/android/inbox/IncomingMessageDispatcher.kt:431-477`: receive path has no upper bound on `payload.avatar.size`. Sender pipeline caps at 16 KB (`GroupAvatarImage.MAX_BYTES`), but a misbehaving/forked admin client could ship a larger blob and it'd be persisted unchecked. Relay event-size limits effectively bound this, but adding a defensive `payload.avatar?.size ?: 0 <= GroupAvatarImage.MAX_BYTES * 2` reject would harden the receive gate and is symmetric with the sender contract.
- `app/src/main/kotlin/app/onym/android/group/GroupAvatarBroadcaster.kt:60-67`: admin gate matches BLS-hex case-insensitively (`adminPubkeyHex?.lowercase()` vs `activeBlsHex`), but the dispatcher's receive-side gate (`IncomingMessageDispatcher.kt:457-459`) compares Ed25519-hex via `sender.toHexLowercase() != storedAdmin.lowercase()` — note the broadcaster gates on BLS, the receiver on Ed25519. That's fine because the envelope sealer ties the two together, but the asymmetry is worth a one-line comment to prevent future "tighten the gate" patches from drifting them apart.
- `app/src/main/kotlin/app/onym/android/chats/ChatMembersScreen.kt:115`: `canEditAvatar = canShareInvite` works today because both are admin-gated, but coupling avatar-edit to "can share invite" semantically is fragile — if Share-Invite ever opens to non-admins (or the variable's meaning shifts), avatar edit silently follows. Consider a dedicated `isAdmin` derivation reused by both, or at least a comment binding them.
- `app/src/main/kotlin/app/onym/android/chats/ChatsViewModel.kt:54-58`: `setGroupAvatar` swallows `broadcaster.setAvatar`'s `Outcome`. UI has no signal for `UnknownGroup` / `NotAdmin` / `NoIdentity` — picker silently no-ops. Probably fine for the admin-only happy path, but worth noting since the broadcaster carefully returns four distinct outcomes that are then discarded.
- Tests are thorough and mirror the iOS twin: wire-format pin (`GroupAvatarPayloadTest`), encrypted-at-rest column (`RoomGroupStoreTest`), trust gate + idempotency + unknown-group + admin-less paths (`IncomingMessageDispatcherAvatarTest`), fan-out skip-self + non-admin rejection + removal-omits-key (`GroupAvatarBroadcasterTest`), and real-JPEG budget check (instrumented `GroupAvatarImageTest`). Coverage matches the surface area.

dev-onym (commented):
APPROVE — solid iOS-parity implementation with strong tests and well-reasoned trust/wire choices. A few non-blocking notes:

- `GroupAvatarBroadcaster.kt:74-80` — read-modify-write race: the broadcaster snapshots `group` then later `insert(group.copy(avatar = avatar))`. A concurrent membership change (join approval, profile update) landing between the two could be clobbered by the stale snapshot. Low likelihood for an admin's own photo flow, but worth a thought (or a repository-level merge).
- `GroupAvatarBroadcaster.kt:79-86` — encode-failure path returns `Outcome.Sent` even though no fan-out happened ("Local apply already succeeded; skip the fan-out"). Caller can't distinguish "fanned out" from "applied locally only" — consider a separate outcome (or at least log), otherwise the receipt is misleading.
- `ChatMembersScreen.kt:367-371` — `remember(group.avatar)` keys on the `ByteArray` identity, not content; a fresh snapshot with byte-equal avatar triggers a re-decode on every recomposition. Cheap, but `remember(group.avatar?.contentHashCode())` would avoid it. Also, neither the picked-image bitmap (`decodeAndEncodeAvatar`) nor the header bitmap is `recycle()`d — small-image, but consider on low-mem devices.
- `GroupAvatarBroadcaster.kt:84` — `catch (_: Throwable)` around `encodeToString` is broader than needed; serialization errors are `SerializationException`. Same pattern on `sealInvitation` at line 95. Minor — `Throwable` swallows `CancellationException`, which can mask coroutine cancellation in the broadcast loop.
- `IncomingMessageDispatcher.kt:457-471` — idempotency check is correct, but consider hoisting the `tryDecodeGroupAvatar` decode to also short-circuit malformed payloads earlier; right now any JSON-shaped blob gets a full decode attempt for every inbound message (chat-message branch was previously first). Throughput impact is negligible for normal traffic, just noting the ordering cost.
- `GroupAvatarPayloadTest.kt:78` — `assertTrue("removal must omit avatar key", !text.contains("\"avatar\":"))` would also fail-positive if any other key starts with `"avatar"` — but `senderBlsHex`/`sentAtMillis`/etc. are all prefixed `avatar_`, so `"avatar":` (with the colon and no underscore) uniquely matches the field. Fine, but brittle to future field renames.
- Migration `MIGRATION_5_6` (`GroupDatabase.kt:62-70`) is non-destructive `ALTER TABLE ADD COLUMN BLOB` — looks correct and matches the v4→v5 pattern.

PR #171, opened 2026-05-23
feat(groups): invite by QR — show from Group Settings, scan to join

lead-onym:
Reaches parity with iOS PR onymchat/onym-ios#168 — "invite by QR code" for group invites.

### Host side — show the invite as a QR
- `ShareInviteScreen` (reached from Group Settings → members roster via the admin-only "Share invite" action, and reused as the post-create invite screen) now renders the minted invite link as a QR code above the existing Share / Copy buttons, with a caption: *"Have them scan this with their camera to join."*
- Reuses the existing **`OnymQrCode`** component (the same brand-anchored QR the identity invite-key screen uses) — no new generator.
- Encodes the **exact** `s.link` string the Share/Copy buttons already hand out (`https://onym.app/join?c=<…>`), untruncated.

### Joiner side — scan to join, in-app
- New **`ScanToJoinScreen`** wraps the existing generic **`QrScannerScreen`** (camera + permission flow reused as-is) and runs whatever it decodes through the **same** `DeeplinkCapture.introCapabilityFromUri` allowlist + `IntroCapability` decoder the tapped-link path uses. A scanned QR and a tapped link reach the identical `join_invite/{capability}` destination — no second decoder, no second join path.
- A scanned string that isn't an Onym invite (decoder returns `null`) shows a **"Not an Onym invite"** snackbar and re-arms the camera instead of starting a join.
- Two entry points on the Chats screen:
  - an always-visible **QR scanner action in the top bar**, and
  - a secondary **"Scan to join"** button in the **empty state** (a brand-new user joining their first group has no chat yet).
- Strings added in `en` + `ru`.

## Test plan
- [ ] **Host shows QR:** open a Tyranny group → members roster → Share invite → a QR renders above Share/Copy; the QR encodes the same link the Copy button puts on the clipboard.
- [ ] **Post-create:** create a group → the share screen shows the same QR.
- [ ] **Scan valid invite:** on a second device/identity, Chats → QR action (or empty-state "Scan to join") → scan the host's QR → lands on the Join screen for that group (same as tapping the link).
- [ ] **Scan non-Onym QR:** scan an arbitrary QR (e.g. a plain URL) → "Not an Onym invite" snackbar, camera stays live, no join started.
- [ ] **Tapped link still works:** opening `https://onym.app/join?c=…` / `onym://join?c=…` still routes to Join unchanged.

## ⚠️ Cross-platform blocker (flagged, not fixed here)
iOS found that the recent "avatar in the group invitation snapshot" work can break joining on **both** scan and tapped-link paths. **Android has the same issue, confirmed:**

- `GroupInvitationPayload.avatar` (`app/.../group/GroupInvitationPayload.kt`) carries the raw group-avatar **JPEG** (budget `GroupAvatarImage.MAX_BYTES = 16 KB` raw → ~22 KB base64 on the wire), and `CreateGroupInteractor.sendInvitations` includes it in the sealed invitation.
- The sealed envelope is then **base64-encoded again** into the Nostr event `content` (`NostrInboxTransport.buildSendEvent`), roughly doubling size.
- There is **no pre-publish size check**, and `NostrRelayConnection.publishAndAwaitOK` **treats a publish timeout as success**; the create path only checks `acceptedBy >= 1`. So a strict relay that drops the oversized event silently is indistinguishable from success, and the joiner never materializes the group.

This is independent of the QR work — it would break joining regardless. Recommend tracking + fixing separately (e.g. move the avatar out of the invitation snapshot onto the dedicated `GroupAvatarPayload` channel, and/or add a pre-publish size guard + distinguish `OK false` from timeout).

### DIFF
--- a/app/src/main/kotlin/app/onym/android/RootScreen.kt
+++ b/app/src/main/kotlin/app/onym/android/RootScreen.kt
@@ -37,6 +37,7 @@ import app.onym.android.group.CreateGroupViewModel
 import app.onym.android.group.IntroCapability
 import app.onym.android.group.JoinScreen
 import app.onym.android.group.JoinViewModel
+import app.onym.android.group.ScanToJoinScreen
 import app.onym.android.group.ShareInviteViewModel
 import app.onym.android.group.creategroup.CreateGroupScreen
 import app.onym.android.group.creategroup.ShareInviteScreen
@@ -189,6 +190,21 @@ fun RootScreen(
                         // tap deeper behind the thread's info button.
                         navController.navigate("chat_thread/$groupId")
                     },
+                    onScanToJoin = { navController.navigate(ROUTE_SCAN_JOIN) },
+                )
+            }
+            composable(ROUTE_SCAN_JOIN) {
+                ScanToJoinScreen(
+                    onCapability = { capability ->
+                        // Route a scanned invite through the exact join
+                        // destination the tapped-link path uses; drop the
+                        // scanner from the back stack so Back from Join
+                        // returns to Chats, not the live camera.
+                        navController.navigate("join_invite/${capability.encode()}") {
+                            popUpTo(ROUTE_SCAN_JOIN) { inclusive = true }
+                        }
+                    },
+                    onCancel = { navController.popBackStack() },
                 )
             }
             composable("chat_thread/{groupId}") { entry ->
@@ -566,4 +582,5 @@ private const val ROUTE_ANCHORS_ROOT = "anchors_root"
 private const val ROUTE_CREATE_GROUP = "create_group"
 private const val ROUTE_APPROVE_REQUESTS = "approve_requests"
 private const val ROUTE_PENDING_INVITES = "pending_invites"
+private const val ROUTE_SCAN_JOIN = "scan_join"
 private val TAB_ROUTES = setOf(Tab.Chats.route, Tab.Settings.route, Tab.Search.route)
--- a/app/src/main/kotlin/app/onym/android/chats/ChatsScreen.kt
+++ b/app/src/main/kotlin/app/onym/android/chats/ChatsScreen.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Forum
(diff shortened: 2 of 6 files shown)


dev-onym (commented):
APPROVE

- `ScanToJoinScreen` re-arms via `key(attempt)` (`group/ScanToJoinScreen.kt:57`), which on each non-Onym scan tears down CameraX entirely (`scan/QrScannerScreen.kt:232-238` onDispose unbinds + shuts down the analysis executor) and re-creates the `PreviewView`/binding. Works (permission is preserved, so no re-prompt) but causes a visible camera reinit flicker per rejected scan. Lighter alternative: expose an `armKey: Any` param on `QrScannerScreen` that only resets the `consumed` latch (`scan/QrScannerScreen.kt:189`) without rebuilding camera state. Not blocking — rejected scans are rare.
- `EmptyState` secondary affordance uses `titleMedium` and an 18.dp icon (`chats/ChatsScreen.kt:262-282`) — same text style as the primary "Create group" CTA above it, so the visual hierarchy doesn't read as "secondary" despite the comment claiming so. Consider `bodyMedium` or `labelLarge`. Minor.
- Snackbar `scope` is `rememberCoroutineScope()` on the outer `ScanToJoinScreen` (`group/ScanToJoinScreen.kt:50`) — survives the inner `key(attempt)` re-key, so the "Not an Onym invite" snackbar reliably shows. Good.
- `DeeplinkCapture.introCapabilityFromUri(rawUri: String?)` (`transport/DeeplinkCapture.kt:49`) accepts the raw scanned string directly — same decoder as the tapped-link path (`MainActivity.kt:90, 102`). The "no second decoder" claim holds.
- `ShareInviteScreen.kt:125-140` `OnymQrCode(value = s.link, …)` encodes the exact `s.link` (`group/creategroup/ShareInviteScreen.kt:125`), matching what Share/Copy hand out. `TextAlign`, `Spacer`, `OnymQrCode` imports all present.
- `RootScreen.kt:194-205` `popUpTo(ROUTE_SCAN_JOIN) { inclusive = true }` correctly removes the live-camera entry from the back stack on capability hit — Back from Join returns to Chats, not the scanner.
- Strings added in both `values/strings.xml:402,441-442` and `values-ru/strings.xml:394,433-434`. Parity OK.
- Cross-platform blocker (oversized invite payload, no pre-publish size check, publish-timeout-as-success) is correctly scoped as out-of-band of this PR; tracking it separately is the right call.

PR #172, opened 2026-05-23
fix(inbox): stop the launch-time get_commitment storm that breaks live joins

lead-onym:
Ports iOS PR onymchat/onym-ios#169 to Android. The Kotlin receive-side had the same bug.

## Symptom
A joiner who scans/taps an invite and is approved often **doesn't see the group live** — it only appears after an app restart. The invitation arrives and persists fine (hence restart works); the live materialization is being dropped.

## Root cause (relayer-side, not transport)
- Inbox (Nostr) subscriptions use **no `since` filter** (`NostrInboxTransport.subscriptionFilters`), so every relay reconnect replays the **entire inbox history**.
- The dispatcher did an **uncached `get_commitment`** per replayed Tyranny group-state message (`verifyTyrannyInvitation`, `verifyTyrannyAnnouncement`) — and **before any dedup** (`applyAnnouncement`'s member-key check ran *after* the chain read).
- That self-inflicted burst **throttled the relayer**. A throttled/lagging read was then treated as a **hard reject**, silently dropping the very invitation being verified.
- The same hard-reject hit a legitimate **chain-behind** read: the admin anchors epoch N then immediately sends the snapshot; the joiner's relayer read still lags at N-1 → `onchain.epoch < snapshot.epoch` → reject + drop.

## Fixes (mirror the four in the iOS PR)
1. **Skip the chain read when already locally verified.** `verifyTyrannyInvitation` returns `VERIFIED` without a relayer call if a group with the exact `(commitment, epoch)` is already materialized. The local Poseidon recompute still runs **first**, so a replay that swaps the roster while reusing the commitment is still rejected. This removes nearly all of the startup reads.
2. **Dedup announcements before the chain read.** Moved the BLS-pubkey-hex dedup in `applyAnnouncement` *above* `verifyTyrannyAnnouncement`, so re-delivered known members don't each hit the relayer.
3. **Throttle/lag ⇒ defer, not reject.** A chain-read throw (throttled/offline/unconfigured) and a chain-behind read now return `STALE_NEEDS_REFRESH` (defer + admin-refresh retry via `GroupStateVerifier.deferVerification`) instead of `REJECT`. **Deferral never materializes without a later exact-epoch chain match**, so forgery protection is unchanged — only the false-drop is fixed.
4. **`CachingChainStateReader`** — bounded retry (linear backoff) + short (10s) TTL cache around the raw relayer reads, wired in `OnymApplication`. `SepContractChainStateReader` stays cache-free. A stale cache entry can only ever cause a brief *deferral* (never a false reject, given #3), so caching is safe here.

## Why it's safe (security)
The only `REJECT` paths that remain are genuine cryptographic failures: **missing commitment**, **internal recompute mismatch**, and **commitment mismatch at an exact epoch**. All "couldn't verify *right now*" cases defer and re-verify, and never materialize a group without an exact-epoch on-chain match.

## Test plan
**JVM unit tests (run, green):**
- `CachingChainStateReaderTest` — cache hit within TTL (inner called once), expiry (called again), retry recovers from N transient failures, retry exhausted rethrows.
- `IncomingMessageDispatcherTest.announcement_tyranny_knownMember_skipsChainRead` — known-member announcement replay → chain reader called **0** times.
- All pre-existing dispatcher reject/materialize/converge-forward tests still green.

**FFI tests (compile clean; require an emulator/device to run, like all `*FfiTest`):** added to `IncomingMessageDispatcherConvergeForwardFfiTest` (real Poseidon recompute needs the native `.so`):
- redelivered already-verified invitation → chain reader called **exactly once** across two deliveries.
- chain-read throw → **defers** (SpyRefresher records the groupId), does not materialize.
- chain-behind → **defers**, does not materialize.
- existing exact-epoch-materializes and chain-ahead-defers tests unchanged.

```
./gradlew :app:testDebugUnitTest           # green
./gradlew :app:compileDebugAndroidTestKotlin  # green
```

## Follow-up (not in this PR)
Cap the Nostr replay itself — a `since` / processed-event high-water-mark on the inbox subscription — to cut inbound bandwidth. This PR targets the **relayer RPM**, which was the harmful part (the throttling that dropped joins). The replay cap is a bandwidth optimization on top.

### DIFF
--- a/app/src/androidTest/kotlin/app/onym/android/inbox/IncomingMessageDispatcherConvergeForwardFfiTest.kt
+++ b/app/src/androidTest/kotlin/app/onym/android/inbox/IncomingMessageDispatcherConvergeForwardFfiTest.kt
@@ -104,6 +104,112 @@ class IncomingMessageDispatcherConvergeForwardFfiTest {
         )
     }
 
+    @Test
+    fun invitation_tyranny_redelivery_skipsChainReadWhenAlreadyVerified() = runBlocking {
+        // The launch-time storm fix: a re-delivered invitation for an
+        // already-materialized (commitment, epoch) must NOT hit the relayer
+        // again. First delivery verifies (1 chain read) and materializes;
+        // the identical replay short-circuits on the local match, leaving
+        // the chain-read count at 1.
+        val env = environment()
+        val realCommitment = realCommitment(epoch = 0uL)
+        val plaintext = encode(invitation(commitment = realCommitment, epoch = 0uL))
+        val counting = CountingChainState(SepCommitmentEntry(commitment = realCommitment, epoch = 0uL))
+        val dispatcher = IncomingMessageDispatcher(
+            envelopeDecrypter = FakeInvitationEnvelopeDecrypter(
+                FakeInvitationEnvelopeDecrypter.Mode.Fixed(plaintext),
+            ),
+            groupRepository = env.groups,
+            invitationsRepository = env.invitations,
+            chainState = counting,
+            groupStateRefresher = SpyRefresher(),
+        )
+
+        dispatcher.dispatch("mat-1", owner, byteArrayOf(), Instant.EPOCH)
+        dispatcher.dispatch("mat-1-replay", owner, byteArrayOf(), Instant.EPOCH)
+
+        assertEquals(
+            "idempotent — still exactly one group",
+            1,
+            env.groups.snapshots.value.count { it.groupIdBytes.contentEquals(groupId) },
+        )
+        assertEquals(
+            "replay of an already-verified snapshot must not re-read the chain",
+            1,
+            counting.callCount,
+        )
+    }
+
+    @Test
+    fun invitation_tyranny_chainReadThrows_defersInsteadOfReject() = runBlocking {
+        // A throttled / unreachable relayer is not evidence of forgery. The
+        // invitation must be deferred (retried via the admin-refresh path),
+        // never silently dropped — that drop was the root cause of "joiner
+        // only sees the chat after a restart".
+        val env = environment()
+        val realCommitment = realCommitment(epoch = 0uL)
+        val plaintext = encode(invitation(commitment = realCommitment, epoch = 0uL))
+        val refresher = SpyRefresher()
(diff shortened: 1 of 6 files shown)


dev-onym (commented):
I've reviewed the diff and the surrounding flow in `IncomingMessageDispatcher.kt`, `GroupStateVerifier.kt`, and the new `CachingChainStateReader.kt`. The three claims I checked all hold: STALE_NEEDS_REFRESH is wired to `deferVerification` (IncomingMessageDispatcher.kt:236-239), defer cannot materialize without a fresh exact-epoch verification (GroupStateVerifier.kt:216-226 + the admin-reply re-entering the dispatcher), and Poseidon recompute runs before the local-match short-circuit (IncomingMessageDispatcher.kt:621-639).

APPROVE — fix is correct and security argument holds; a few minor non-blockers.

- `CachingChainStateReader.kt:48-49` — cache is unbounded (no max size / LRU eviction). Fine in practice (groupId-keyed, bounded by groups joined), but worth a note.
- `CachingChainStateReader.kt:60-71` — retry is unconditional: even non-transient errors (e.g. `ChainReadError.NoContractBinding`, exercised by `retry_exhausted_rethrows`) burn 3 attempts + 900ms backoff. Consider classifying transient vs permanent.
- `CachingChainStateReader.kt:60-71` — concurrent callers for the same groupId all miss together (no in-flight coalescing) and each fire `inner`; a thundering herd on cold start could partially defeat the storm-suppression goal. Mutex only guards the map, not the fetch.
- `CachingChainStateReader.kt:73-75` / `GroupStateVerifier.kt:277-278` — identical `toHexLowercase` helper duplicated; minor.
- `IncomingMessageDispatcher.kt:391-399` — dedup-before-trust-check is a behavior change: a re-announcement can no longer update a known member's `alias`/`sendingPubkey`/`inboxPub`. Pre-PR behavior overwrote them on every replay. The PR body doesn't call this out — likely intended (matches iOS), but confirm no flow relied on announcement-driven profile updates.
- Tests look thorough and match the PR's claims. Cache positive-only (no negative caching of failures) is the right call given the defer-on-error semantics.

PR #174, opened 2026-05-27
feat(chat): message replies — swipe to reply, inset quote, tap to jump

lead-onym:
Telegram-style replies in chat, wire-compatible with iOS (#173–#175). Swipe a message to reply; a banner shows what you're replying to; the sent message renders an inset quote (sender + snippet) above its text; tapping the quote jumps to the original.

## Wire contract (matches iOS exactly)
The chat payload gains one optional field: `reply_to_message_id` (snake_case) — the target message's UUID string, or absent/null for non-replies.

- **Additive + optional, so `version` stays `1`.** An older sender omits the key (decodes to null); an older receiver ignores it.
- Only the target id travels. The quoted sender/body are **not** snapshotted — the receiver resolves the quote live from its own local store at render time.
- No new trust gate on receive: a dangling/forged ref just renders "Message unavailable" (rendering only resolves locally-known messages).
- Replies ride the existing Tyranny-only chat path; other governance types stay unsupported.

## Changes

**Data layer** (iOS #173)
- `ChatMessagePayload` — optional `reply_to_message_id` (uppercase canonical UUID, same encoding as `message_id`).
- `ChatMessage` / `PersistedMessage` — plain (non-encrypted) reply pointer; it's just a row reference, left queryable.
- `MessageDatabase` v1 → v2 with a non-destructive `ALTER TABLE … ADD COLUMN` migration (wired ahead of the existing destructive fallback, so existing messages survive).
- `SendMessageInteractor.send` takes the ref (default null) → payload + optimistic row; `retry` reconstructs it from the stored row. `IncomingMessageDispatcher` maps the ref onto the inbound message.

**Render** (iOS #174)
- `ChatReplyQuote` + live resolution from the loaded list: found → quote with the target sender's name + accent + one-line snippet; missing → muted "Message unavailable", not tappable.
- Quote renders above the bubble body; tapping it scrolls to and briefly flashes the original.

**Compose** (iOS #175)
- Drag-to-reply on the bubble: it follows the finger leftward (clamped), a reply glyph grows in, a one-shot haptic fires at the threshold, and it springs back; releasing past the threshold arms the reply. Leftward-only so vertical scroll and the back-swipe are untouched.
- "Replying to {name}" banner above the composer with a cancel button; arming focuses the composer. The target threads through the VM's send and clears on send / cancel.

Strings ("Replying to {name}", "Message", "Message unavailable", "Cancel reply") are plain literals, matching the existing chat UI and the iOS choice.

## Tests
Payload round-trip + back-compat decode (missing / explicit-null key → null) + snake_case pin; store round-trip; inbound mapping carries the ref; send carries it on payload + persisted row and retry preserves it; quote resolution (found / missing / non-reply / fingerprint fallback); and VM arm / cancel / send-forwards-and-clears. Full `:app:testDebugUnitTest` passes.

### DIFF
--- a/app/src/main/kotlin/app/onym/android/AppDependencies.kt
+++ b/app/src/main/kotlin/app/onym/android/AppDependencies.kt
@@ -42,6 +42,9 @@ class AppDependencies(
     /** App-wide testnet/mainnet preference. Settings exposes a Switch
      *  bound to this; CreateGroupInteractor reads it per call. */
     val networkPreferenceProvider: NetworkPreferenceProvider,
+    /** Symmetric read-receipt setting, bound to the Settings → Chat
+     *  toggle and read by the dispatcher + chat thread. */
+    val readReceiptsPreferenceProvider: app.onym.android.chats.ReadReceiptsPreferenceProvider,
     val makeCreateGroupViewModel: () -> CreateGroupViewModel,
     /** Chats tab — read-only view over [app.onym.android.group.GroupRepository.snapshots].
      *  Mirrors `makeChatsFlow` from onym-ios PR #30. */
--- a/app/src/main/kotlin/app/onym/android/OnymApplication.kt
+++ b/app/src/main/kotlin/app/onym/android/OnymApplication.kt
@@ -323,6 +323,9 @@ class OnymApplication : Application() {
                 app.onym.android.chats.MessageDatabase::class.java,
                 "app.onym.android.messages",
             )
+                .addMigrations(
+                    app.onym.android.chats.MessageDatabaseMigrations.MIGRATION_1_2,
+                )
                 .fallbackToDestructiveMigration()
                 .build()
         } catch (_: Throwable) {
@@ -670,7 +673,9 @@ class OnymApplication : Application() {
                     groupId = groupId,
                     groupRepository = groupRepository,
                     messageRepository = messageRepository,
-                    sendMessage = sender::send,
+                    sendMessage = { gid, body, replyTo ->
+                        sender.send(gid, body, replyTo)
+                    },
                     retryMessage = sender::retry,
                 )
             },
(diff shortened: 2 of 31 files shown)


dev-onym (commented):
COMMENT

- `pointerInput(onSwipeReply)`: ChatBubble.kt:177 keys on a lambda that callers create fresh per recomposition (`onSwipeReply = { onArmReply(message.id) }` at ChatThreadScreen.kt:309). Any parent recomposition during a drag (new message tick, `replyingTo` change, sender-display rebuild) tears down the gesture mid-swipe. Key on `message.id` (stable) instead, e.g. `pointerInput(message.id) { ... onSwipeReply() ... }`.
- Per-event coroutine churn: ChatBubble.kt:203 does `scope.launch { dragOffset.snapTo(dragAccumulated) }` on every drag tick. `snapTo` is itself a suspend, so launching to set a value is fine but allocates a coroutine per pixel of drag. `dragAccumulated` and `swipeArmed` (ChatBubble.kt:120-122) are `mutableStateOf` but only ever read inside the gesture callback — using `remember { Ref... }`/plain holders avoids spurious recompositions.
- Silent focus failure: ChatThreadScreen.kt:337 `runCatching { composerFocus.requestFocus() }` swallows the only signal that the requester isn't attached. If the composer is reachable while a banner shows, this shouldn't throw — drop the `runCatching` and let a real failure surface.
- Wire-pin coverage: the new `wireFormat_replyKeyIsSnakeCaseUppercaseUuid` (ChatMessagePayloadTest.kt) only asserts the present case. The existing strict-keys pin should be extended to cover the non-reply branch (does Kotlin emit `"reply_to_message_id": null` or omit it?) so future serializer-config drift can't silently change the bytes a non-reply ships under v1.
- Migration safety: MessageDatabase.kt:25 bumps to v2 and OnymApplication.kt:326 wires `MIGRATION_1_2` *before* `fallbackToDestructiveMigration` — good. No instrumented test exercises the v1→v2 path; consider one before merging since the fallback would silently nuke prod chats if the ALTER ever regresses.
- Plaintext reply pointer: PersistedMessage.kt:56 stores `replyToMessageId` unencrypted-at-rest by design (documented). Confirms only conversation-graph metadata leaks on device compromise, not body — worth being explicit if threat model treats reply-graph as sensitive.

